Issue #60, opened 2023-01-10
Error exception "Expected keys to be an array of string" on Vercel deployment


obsnomad:
**Expected Behavior**
No error when calling getAll().

**Current Behavior**
Working fine on dev environment, throws "Expected keys to be an array of string" exception on Vercel deployment.

**Possible Solution**
https://github.com/vercel/edge-config/blob/9689e9a0e21bba231e802074096a5a338f83c64f/src/index.ts#L257 should be conditioned like on line 265.

furf:
Was just arriving to call out the same issue. The assertion appears at odds with the functionality! 🥷🤺

https://github.com/vercel/edge-config/blob/main/src/index.ts#L34-L40
https://github.com/vercel/edge-config/blob/main/src/index.ts#L256-L262

Quick suggestion:
``` typescript
        if (localEdgeConfig) {
          if (keys !== undefined) {
            assertIsKeys(keys);
          }

          return Array.isArray(keys)
            ? Promise.resolve(clone(pick(localEdgeConfig.items, keys)) as T)
            : Promise.resolve(clone(localEdgeConfig.items) as T);
        }
```

furf:
Perhaps more accurately:
``` typescript
        if (localEdgeConfig) {
          if (keys === undefined) {
            Promise.resolve(clone(localEdgeConfig.items) as T);
          }

          assertIsKeys(keys);
          return Promise.resolve(clone(pick(localEdgeConfig.items, keys)) as T)
        }
```

dferber90:
Closing this as it shipped in #61

PR #61, opened 2023-01-10
only assert keys in getAll when necessary

furf:
addresses issue #60 where overzealous assertion prevents use of getAll to "get all" 😆


dferber90 (approved):
Thanks for catching and fix this!

PR #64, opened 2023-01-23
improves type generics for `get` and `getAll`

jzxhuang:
I noticed that the type generics for `get` and `getAll` can be improved to be more accurate and restrictive (especially for `getAll`).

First, let's define the following type — edge config is basically a json store, so any value in the edge config must be a valid JSON value.

```ts
type JsonValue =
  | string
  | number
  | boolean
  | null
  | { [x: string]: JsonValue }
  | JsonValue[];

export type EdgeConfigValue = JsonValue;
```
### Previous type definition for `getAll`:

```ts
getAll: <T = any>(keys?: (keyof T)[]) => Promise<T | undefined>;
```

Issues: 
* It's incorrect to allow the response to be any type. We know that `getAll` always returns a record (JSON object), so the return type should be typed as a `Record`. Currently,  calling `getAll<string>` would return a response `string`, which is never possible. The return type should actually be more like `Record<string, string>`.
* We can default `T` to `EdgeConfigValue`.
* If `keys` is defined, we can infer the keys of the Record based on the `keys` that's passed in. 

Examples:
```ts
const a = await getAll()
// a: any
// expected: a: Record<string, JsonValue>

const b = await getAll<string>()
// b: string
// expected: b: Record<string, string>

const c = await getAll(['foo', 'bar'])
// c: any
// expected: c: Partial<Record<'foo' | 'bar', JsonValue>>
```

### New type definition:
```ts
  getAll: <
    T extends EdgeConfigValue = EdgeConfigValue,
    K extends string = string,
  >(
    keys?: K[],
  ) => Promise<Partial<Record<K, T>> | undefined>;
```

Examples: 
```ts
const a = await getAll()
// a: Partial<Record<string, EdgeConfigValue>> | undefined

const b = await getAll<string>()
// b: Partial<Record<string, string>> | undefined

const c = await getAll(['foo', 'bar'])
// c: Partial<Record<'foo' | 'bar', EdgeConfigValue>> | undefined
```

Note: I wrapped the Record in `Partial` because the keys are not guaranteed to exist. For example:
```ts
// edge config: { x: 'x' }

const a = await getAll(['foo']) // returns {}
// a.x = undefined
```

## `get`

Before:
```ts
get: <T = any>(key: string) => Promise<T | undefined>;
```

After:
```ts
  get: <T extends EdgeConfigValue = EdgeConfigValue>(
    key: string,
  ) => Promise<T | undefined>;
```

P.S. I didn't see any contributing guidelines but I just went ahead and made a PR, hopefully y'all don't mind!

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,4 +1,4 @@
-import type { EmbeddedEdgeConfig } from './types';
+import type { EdgeConfigValue, EmbeddedEdgeConfig } from './types';
 
 declare global {
   /* eslint-disable camelcase */
@@ -125,10 +125,15 @@ async function getLocalEdgeConfig(
  * Edge Config Client
  */
 export interface EdgeConfigClient {
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  get: <T = any>(key: string) => Promise<T | undefined>;
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  getAll: <T = any>(keys?: (keyof T)[]) => Promise<T | undefined>;
+  get: <T extends EdgeConfigValue = EdgeConfigValue>(
+    key: string,
+  ) => Promise<T | undefined>;
+  getAll: <
+    T extends EdgeConfigValue = EdgeConfigValue,
+    K extends string = string,
+  >(
+    keys?: K[],
+  ) => Promise<Partial<Record<K, T>> | undefined>;
   has: (key: string) => Promise<boolean>;
   digest: () => Promise<string>;
 }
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -125,10 +125,15 @@ async function getLocalEdgeConfig(
...
-  getAll: <T = any>(keys?: (keyof T)[]) => Promise<T | undefined>;
+  get: <T extends EdgeConfigValue = EdgeConfigValue>(
+    key: string,
+  ) => Promise<T | undefined>;
+  getAll: <
+    T extends EdgeConfigValue = EdgeConfigValue,
+    K extends string = string,
+  >(
+    keys?: K[],
+  ) => Promise<Partial<Record<K, T>> | undefined>;
```
jzxhuang, line 136: one change I didn't make, but I think would be more accurate:
```suggestion
  ) => Promise<Partial<Record<K, T>>>;
```

It doesn't look like it's even possible to make an edge config `undefined` (I tried in the Vercel dashboard, not sure about through the API). So the response would always be at least `{}`.

But I wasn't 100% sure here, so I didn't make this change

dferber90:
Thanks a ton for this detailed writeup. 

I ended up going for a slightly different approach in #67 that's totally based on the work you prepared in this PR. I couldn't merge this PR as we recently split up the `index.ts` file into `index.node.ts` and `index.edge.ts` so it had merge conflicts.

I gave you a shout-out in the changelog and hope that makes up for it :)

PR #67, opened 2023-02-07
introduce EdgeConfigValue type

dferber90:
kudos to @jzxhuang for drafting this in #64

- based on https://github.com/vercel/edge-config/pull/64
- used a fresh branch as #64 has merge conflicts and as I went for a slightly different approach

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -1,17 +1,27 @@
 export interface EmbeddedEdgeConfig {
   digest: string;
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  items: Record<string, any>;
+  items: Record<string, EdgeConfigValue>;
 }
 
 /**
  * Edge Config Client
  */
 export interface EdgeConfigClient {
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  get: <T = any>(key: string) => Promise<T | undefined>;
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  getAll: <T = any>(keys?: (keyof T)[]) => Promise<T | undefined>;
+  get: <T extends EdgeConfigValue = EdgeConfigValue>(
+    key: string,
+  ) => Promise<T | undefined>;
+  getAll: <T extends EdgeConfigItems = EdgeConfigItems>(
+    keys?: (keyof T)[],
+  ) => Promise<T>;
   has: (key: string) => Promise<boolean>;
   digest: () => Promise<string>;
 }
+
+export type EdgeConfigItems = Record<string, EdgeConfigValue>;
+export type EdgeConfigValue =
+  | string
+  | number
+  | boolean
+  | null
+  | { [x: string]: EdgeConfigValue }
+  | EdgeConfigValue[];
--- a/.changeset/big-garlics-protect.md
+++ b/.changeset/big-garlics-protect.md
@@ -1,0 +1,5 @@
+---
+'@vercel/edge-config': patch
+---
+
+make `get()` and `getAll()` return type EdgeConfigValue instead of any, kudos @jzxhuang
--- a/src/index.edge.ts
+++ b/src/index.edge.ts
@@ -4,24 +4,21 @@ import {
   parseConnectionString,
   ERRORS,
 } from './shared';
+import type {
+  EdgeConfigClient,
+  EdgeConfigItems,
+  EdgeConfigValue,
+  EmbeddedEdgeConfig,
+} from './types';
 
-export { parseConnectionString };
+export {
+  parseConnectionString,
+  type EdgeConfigClient,
+  type EdgeConfigItems,
+  type EdgeConfigValue,
+  type EmbeddedEdgeConfig,
+};
 
-/**
- * Edge Config Client
- */
-export interface EdgeConfigClient {
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  get: <T = any>(key: string) => Promise<T | undefined>;
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  getAll: <T = any>(keys?: (keyof T)[]) => Promise<T | undefined>;
-  has: (key: string) => Promise<boolean>;
-  digest: () => Promise<string>;
-}
-
-/**
- * Creates a deep clone of an object.
- */
 export function createClient(
   connectionString: string | undefined,
 ): EdgeConfigClient {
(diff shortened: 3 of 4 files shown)


dglsparsons (approved):

src/types.ts
```diff
@@ -1,17 +1,27 @@
...
   digest: () => Promise<string>;
 }
+
+export type EdgeConfigItems = Record<string, EdgeConfigValue>;
+export type EdgeConfigValue =
+  | string
+  | number
+  | boolean
+  | null
+  | { [x: string]: EdgeConfigValue }
```
dglsparsons, line 26: should maybe be a Record<string, EdgeConfigValue>?

dferber90: I'm not a typescript wizard by any means but it seems like it can't be one?

image


https://www.typescriptlang.org/play?#code/C4TwDgpgBAogJgcwgYQPYDsBmBLBA1AQwBsBXCAQSgF4AoKKAHygGdgAnbdBOxqdEgLYAjCGx5MhqVEQgF04viSJEFAbygBtAB4AuFu04IAunvhI0WXIVIUoAXxqOnT0JFiIUGHPmJkAQtQKrBxcCvzCogqS0rLy9Ez8ygoAShAAxqhscAA8wYYANO7mXla+EH4AfEA

jzxhuang:
@dferber90 Thanks for taking a look and getting this merged in quickly! Love the work y'all are doing on this, Edge Config is awesome

One note on this implementation that I think can be improved: As-is, if we write `const result = get(['x', 'y'])`, you'll get the following type for `result`:

```tsx
// typeof result = EdgeConfigItems = Record<string, EdgeConfigValue>

const { z } = result // TS will not complain, since the result is typed as a Record with key `string`
```

But actually, if the user passes in `keys`, then we can (should) restrict the return type to:
```tsx
// typeof result = Partial<Record<K, EdgeConfigValue>>
// where `K` is a key in `keys`

const { z } = result // TS will complain! `z` is not in the array `['x', 'y']`. it is _always_ undefined
const foo = x.bar // TS will complain! x is not guaranteed to be defined
```

Then we can help prevent the consumer from accidentally trying to access `result.z` — and hopefully catch some simple bugs via the type system! 🙈 

I figured out that you can basically express this in TS as the type shown in https://github.com/vercel/edge-config/pull/67/files#diff-c54113cf61ec99691748a3890bfbeb00e10efb3f0a76f03a0fd9ec49072e410aR13-R15. It's kind of verbose so there might be a better way of expressing it, but that's how I figured out how to do it — I'm also not a TS wizard so there might be a more concise way? 🤷 

---

Secondly, and definitely more a style/nit thing but still worth mentioning: I think it's better to make the generic `T` on `getAll` extend just `EdgeConfigValue` instead of `EdgeConfigItems`. Consider:

```tsx
// as-is (T extends EdgeConfigItems)
const moreVerbose = await getAll<Record<string, boolean>>()

// suggested (T extends EdgeConfigValue)
const lessVerbose = await getAll<boolean>()
```

Also, as-is, you could actually write:
```tsx
const result = await getAll<Record<number, boolean>>()
// should be impossible, the key has to be a string I'm pretty sure.
```

I mean, I'm not sure why anyone would write that... but it _is_ possible, and if we can prevent it completely via types, might as well! Mostly, it's just much nicer to write `getAll<{ foo: string, bar: string }>`, instead of`getAll<Record<string, { foo: string, bar: string }>>`. 

We know `getAll` will always return a `Record` with key string — so we only need to make it generic on the value in the `Record`. 

---

Lmk what y'all think — I'm happy to make a PR for this, or I guess it will just be a few simple type changes. Anyways thanks again for the work on this 🙏

dferber90:
Thanks for this input @jzxhuang <picture data-single-emoji=":thumbs:" title=":thumbs:">:thumbs:</picture> 

> Secondly, and definitely more a style/nit thing but still worth mentioning: I think it's better to make the generic T on getAll extend just EdgeConfigValue instead of EdgeConfigItems

` await getAll<boolean>()` This assumes that all Edge Config Items are of the same value, but each key could be holding a value of a different type.

> We know getAll will always return a Record with key string — so we only need to make it generic on the value in the Record.

It expect people to have one type for everything stored in their Edge Config like

```
interface EdgeConfig {
  key1: string;
  key2: number;
}
```

It's very convenient if you can then do `getAll<EdgeConfig>()` or `getAll<EdgeConfig>(["key1", "key2"])`.

In `getAll<EdgeConfig>(["key1", "key2"])` there's a decision about the generic:
- should it specify the return type
- or should `getAll` itself pick by the keys you passed in and make that the return type

src/types.ts
dferber90: Just realized the changes I made prevent interfaces to be passed as the generic now

image


Sample code

```ts
type EdgeConfigType = {
  a: string;
  b: number;
};


interface EdgeConfigInterface {
  a: string;
  b: number;
};

async function main() {
  const _allT = await getAll<EdgeConfigType>();
  const _selectedT = await getAll<EdgeConfigType>(["a", "b"]);

  const _allI = await getAll<EdgeConfigInterface>();
  const _selectedI = await getAll<EdgeConfigInterface>(["a", "b"]);

  console.log(_allT, _selectedT);
  console.log(_allI, _selectedI);
}
```

cc @jzxhuang

jzxhuang:
>  await getAll<boolean>() This assumes that all Edge Config Items are of the same value, but each key could be holding a value of a different type.

@dferber90 Yeah good point! I made the assumption that most of the time, the values in the store will have the same shape — but I was only thinking about my use cases so far. Good call

> In getAll<EdgeConfig>(["key1", "key2"]) there's a decision about the generic:
> 
> should it specify the return type
> or should getAll itself pick by the keys you passed in and make that the return type

Ahh... very interesting! Initial thought:
- `getAll<EdgeConfig>("key1", "key2")` should return the generic that you specify (`EdgeConfig`). 
- `getAll(["key1", "key2"])` should construct the return type as `Partial<Record<keyof keys, EdgeConfigValue>>`. 

Would you agree with this too? Also, I wonder if we can do even better in the case of `getAll<EdgeConfig>("key1", "key2")`: Check that the `keys` array is a superset of the keys of `EdgeConfig`. So then TS will complain if you accidentally write `getAll<EdgeConfig>(["keyOne", "keyTwo"]`.

I'm not sure how to express all this via TS generics, it's starting to get quite complicated even just thinking about the many overloads — but... feels like it should possible 😅

PR #77, opened 2023-03-17
use etags and if-none-match

dferber90:
Uses [ETag](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag) and [if-none-match](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-None-Match) headers to avoid transferring the same data over and over again.

This reduces latency and bandwidth usage.

### DIFF
--- a/src/utils/fetch-with-cache.ts
+++ b/src/utils/fetch-with-cache.ts
@@ -1,0 +1,51 @@
+interface CachedResponsePair {
+  etag: string;
+  response: Response;
+}
+
+type FetchOptions = Omit<RequestInit, 'headers'> & { headers?: Headers };
+
+export const cache = new Map<string, CachedResponsePair>();
+
+export async function fetchWithCache(
+  url: string,
+  options: FetchOptions = {},
+): Promise<Response> {
+  const { headers: customHeaders = new Headers(), ...customOptions } = options;
+  const authHeader = customHeaders.get('Authorization');
+  const cacheKey = `${url},${authHeader || ''}`;
+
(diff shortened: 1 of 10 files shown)


AndyBitz (approved):

src/utils/fetch-with-cache.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  if (cachedResponsePair) {
+    const { etag, response: cachedResponse } = cachedResponsePair;
+    const headers = new Headers(customHeaders);
+    headers.set('If-None-Match', etag);
+
+    const res = await fetch(url, { ...customOptions, headers });
+
+    if (res.status === 304) {
+      return cachedResponse.clone();
```
dferber90, line 28: the current implementation sort of works, but it always responds with the full previous response

a neater approach would likely be to mentally see `fetchWithCache` not as a more powerful fetch, but rather as a layer on top of fetch which only exposes the response headers, status code and data (not body)

AndyBitz (approved):

maggch97:
I have a static web page on vercel and I found vercel always sends the same file content although the etag is same.

Will this PR fix the issue for static web page?

I see this PR merged 3 weeks ago, when will this change go to the production environment?

dferber90:
@maggch97 this repository is for the @vercel/edge-config npm package, which is part of our Edge Config product: https://vercel.com/docs/concepts/edge-network/edge-config.

I believe you are asking about simply hosting a static website without using Edge Config at all, right?

PR #91, opened 2023-04-10
read body to avoid memory leaks

ykzts:
undici fetch and node-fetch maintain the connection until the body is read, causing a memory leak if the body is not read[^1][^2].

In this Pull Request, if a 304 is returned or an error is returned, a meaningless res.arrayBuffer (slow because res.json and res.text are parsed together) is executed to read the body.

[^1]: https://github.com/nodejs/undici/blob/v5.21.2/README.md#garbage-collection
[^2]: node-fetch/node-fetch#83

### DIFF
--- a/.changeset/cyan-ducks-beg.md
+++ b/.changeset/cyan-ducks-beg.md
@@ -1,0 +1,5 @@
+---
+'@vercel/edge-config': patch
+---
+
+read body to avoid memory leaks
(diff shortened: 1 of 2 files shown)


dferber90 (approved):

dferber90:
This is simply brilliant, thank you a lot for bringing this to our attention and fixing it <picture data-single-emoji=":10-10:" title=":10-10:"></picture>

dferber90:
I'm asking around internally to see if we need to apply a similar workaround in `index.edge.ts` which runs in our Edge Functions / Edge Runtime and I'll  release v0.1.7 once that's clarified.

PR #103, opened 2023-05-02
chore: remove `urlpattern-polyfill` from jest setup.js

yusukebe:
It appears that the use of URLPattern is unnecessary, URLPattern has not been used since this commit: https://github.com/vercel/storage/commit/0c78c58412b73640df98f9eec6122dd22ee6b0dc.

### DIFF
--- a/packages/edge-config/jest/setup.js
+++ b/packages/edge-config/jest/setup.js
@@ -1,4 +1,3 @@
 require('jest-fetch-mock').enableMocks();
-require('urlpattern-polyfill');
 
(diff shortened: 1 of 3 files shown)


yusukebe:
This PR will not result in a version bump, so I won't be adding a changeset.

elliott-with-the-longest-name-on-github (approved):
```
Lockout
.
Good-heartedly
.
Take
.
Messy
.
```

Thanks for the sharp eye!

PR #105, opened 2023-05-02
feat: Add integration testing

elliott-with-the-longest-name-on-github:
Adds integration tests for all packages in this app.

TODO:
- [x] README update for Next app and root
- [x] Add environment variables necessary for tests (we need to be able to connect to the pg database)
- [ ] Fast-follow -- add Curated Tests project to this repository (can't do it without `packages/integration` on main)


packages/integration/README.md
```diff
@@ -0,0 +1,11 @@
+Instructions:
+
+- `cd packages/integration`
+- `vc link` => link to `Vercel Labs` => `vercel-storage-integration-test-suite`
```
dferber90: we have a curated-tests team that we should probably use here

packages/integration/.eslintrc.js
```diff
@@ -0,0 +1,30 @@
+module.exports = {
```
dferber90: should we organize the folders differently?

I'd love if we had `packages` as the actual storage packages and other folders for supporting packages, examples and integration tests

That way people checking out this repo on github won't get overwhelmed, and it's also easier for us to see which of these are actually published to npm

elliott-with-the-longest-name-on-github: How about:
- `packages` - maps `@vercel/{name}` to `packages/{name}`
- `tooling` - eslint + tsconfig + others as necessary
- `test` - integration tests

Probably good to put these in `test/next` anyway -- at some point we should add similar tests for a SvelteKit (or other Vite-based framework) app -- those would've caught our issue with the `Proxy`-as-default-export before we went live.

dferber90: Sounds great

elliott-with-the-longest-name-on-github: done

dferber90 (approved):

PR #118, opened 2023-05-04
feat: Allow localhost connection strings

elliott-with-the-longest-name-on-github:
If a connection string is on `localhost`, don't check for whether it's pooled or non-pooled. Just allow it.


adriancooney (approved):
Looks good!

packages/postgres/src/postgres-connection-string.ts
```diff
@@ -37,3 +37,29 @@ export function isPooledConnectionString(connectionString: string): boolean {
 export function isDirectConnectionString(connectionString: string): boolean {
   return !isPooledConnectionString(connectionString);
 }
+
+export function isLocalhostConnectionString(connectionString: string): boolean {
+  try {
+    // This seems silly, but we can use all of the hard work put into URL parsing
+    // if we just convert `postgresql://` to `https://` and then parse it as a URL.
+    const withHttpsProtocol = connectionString.startsWith('postgresql://')
```
klaussner, line 45: I think this should work without replacing the protocol before parsing the URL:
```
> new URL("postgresql://localhost:5432")
URL {
  href: 'postgresql://localhost:5432',
  origin: 'null',
  protocol: 'postgresql:',
  username: '',
  password: '',
  host: 'localhost:5432',
  hostname: 'localhost',
  port: '5432',
  pathname: '',
  search: '',
  searchParams: URLSearchParams {},
  hash: ''
}
```

elliott-with-the-longest-name-on-github: It does not in some cases 😢 Try running the tests without it.

elliott-with-the-longest-name-on-github: image

Actually, this very example doesn't work

klaussner: TIL that [most browsers don’t follow the WHATWG URL spec](https://bugzilla.mozilla.org/show_bug.cgi?id=1374505) when it comes to parsing URLs with unknown protocols ([example](https://felixfbecker.github.io/whatwg-url-custom-host-repro/)). Parsing `postgresql:` connection strings with Node.js should work, though (the tests are all passing). 🤔

PR #122, opened 2023-05-06
Simplify sqlTemplate loop + better guards

WebReflection:
This MR would like to improve the `sqlTemplate` tag in the following way:

  * if the template is valid, in no circumstance a *tag* receives a template with a `length` less than at least one entry
  * if the template is an *array* but any of the following is true, let the driver throw an appropriate and related error
    * the array is empty and `undefined` as string would throw at the driver level
    * the `values` has no index at `${i - 1}`, hence the *tag* was misused and the driver will throw an error
  * there's no need to check `i < values.length` anymore as that doesn't fully guard against possible creation of wrong *SQL* statements (i.e. fake template array and less values actually *could* potentially lead to undesired queries)
  * because of all previous points, there's no need neither to `i + 1` in the loop

### DIFF
--- a/packages/postgres/src/sql-template.ts
+++ b/packages/postgres/src/sql-template.ts
@@ -12,14 +12,11 @@ export function sqlTemplate(
       "It looks like you tried to call `sql` as a function. Make sure to use it as a tagged template.\n\tExample: sql`SELECT * FROM users`, not sql('SELECT * FROM users')",
     );
   }
-  let result = '';
 
-  for (let i = 0; i < strings.length; i++) {
-    result += strings[i];
+  let [result] = strings;
 
-    if (i < values.length) {
-      result += `$${i + 1}`;
-    }
+  for (let i = 1; i < strings.length; i++) {
+    result += `$${i}${strings[i]}`;
   }
 
   return [result, values];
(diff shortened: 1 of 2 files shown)


elliott-with-the-longest-name-on-github (approved):
Thanks!

Tests are failing because forks don't have access to secrets. I verified everything's passing locally and on the deployed test suite.

packages/postgres/src/sql-template.ts
```diff
@@ -12,14 +12,12 @@ export function sqlTemplate(
...
-  for (let i = 0; i < strings.length; i++) {
-    result += strings[i];
+  // ts tries to annotate `strings` with `& any[]` because of the prior check which breaks the type
+  let result = (strings as TemplateStringsArray)[0] ?? '';
 
-    if (i < values.length) {
-      result += `$${i + 1}`;
-    }
+  for (let i = 1; i < strings.length; i++) {
+    result += `$${i}${(strings as TemplateStringsArray)[i] ?? ''}`;
```
WebReflection, line 20: the guard here is `strings.length` already, that `?? ''` shouldn't be necessary but engines here should be smart enough, although I don't understand why TS needs any of these, when `TemplateStringsArray` should have already a `length` greater than `0` by default and also be a list of `string[]` but as this has been merged already, I also don't care much, just wanted to leave my 2c in here.

PR #125, opened 2023-05-07
fix: No more `sql` tomfoolery + CI fixes

elliott-with-the-longest-name-on-github:
This PR contains two minor fixes:
- Live integration tests can run now
- It's basically impossible to call `sql` as a regular function now. You'd have to deliberately pretend to call it like a tagged template by doing something insane like:

```ts
const arr = [];
arr.raw = [];
await sql(arr, []);
```

### DIFF
--- a/.changeset/brown-dodos-sparkle.md
+++ b/.changeset/brown-dodos-sparkle.md
@@ -1,0 +1,6 @@
+---
+'@vercel/postgres': minor
+---
+
+feat: Allow `unsafeUnescaped` insertions into `sql` template calls
+fix: Make it even harder to call `sql` incorrectly
--- a/.changeset/brown-dodos-sparkle.md
+++ b/.changeset/brown-dodos-sparkle.md
@@ -1,6 +1,5 @@
 ---
-'@vercel/postgres': minor
+'@vercel/postgres': patch
 ---
 
-feat: Allow `unsafeUnescaped` insertions into `sql` template calls
 fix: Make it even harder to call `sql` incorrectly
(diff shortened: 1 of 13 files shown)


cramforce (approved):

cprussin:
hey @tcc-sejohnson @cramforce  can you list out some context on why calling `sql` as a regular function was banned?  There's a really nice use for it to construct conditional queries while still benefiting from zeroconf from this lib, e.g. something like:

```ts
const query = (search?: string) => sql(
  `SELECT * FROM someTable ${search ? 'WHERE something LIKE $1' : ''}`,
  search ? [search] : []
);
```

You can of course reach in to the underlying pg client but it's really nice to be able to go through this library for a couple of reasons:

1. zeroconf
2. if this worked, I'd be able to wrap `sql` with my own logic (schema validation, wrapping results in a `Result` type) with a consistent interface regardless of whether the query is conditionally constructed or not

Is there perhaps some other intended API in `@vercel/postgres` for doing this kind of thing?

cramforce:
it was disallowed to avoid accidental SQL injection

cprussin:
> it was disallowed to avoid accidental SQL injection

Got it, thanks @cramforce ! By any chance can you advise on the recommended way to construct queries like the one I posted? I saw `sql.query` but that returns `void` so I can't use it for reads. Do I just have to reach in to pg manually?

cramforce:
Wait, are you sure `sql.query` returns void?

cprussin:
> Wait, are you sure `sql.query` returns void?

That's what typescript is telling me

cramforce:
It returns `Promise<QueryResult<O>>`

cprussin:
huh -- sincere apologies, I just upgraded dependencies and now I'm seeing that too.  Don't know why I was getting `void` before.  Sorry for bothering you and thank you!

PR #129, opened 2023-05-08
feat: Querybuilder API

elliott-with-the-longest-name-on-github:
BREAKING: This adds some functionality to our `sql` API. Prior to this PR, parameterization was limited to places that `pgsql` could accept parameters:

```ts
await sql`SELECT * FROM pokemon WHERE type = ${type};`; // WORKS
await sql`SELECT * FROM ${'pokemon'} WHERE type = ${type};`; // DOES NOT WORK
```

This adds a _very_ simple querybuilder API that allows unsafe, clearly-marked parameterization along with the ability to conditionally append clauses:

```ts
await sql`SELECT * FROM `.appendUnsafeRaw('users').append` WHERE type = ${type};`.execute();
// equivalent to:
await pool.query('SELECT * FROM users WHERE type = $1;', [type]);

// conditionally add clauses, _safely_:
const { type, weight, color } = filters;
const needsWhere = type || weight || color;

const query = sql`SELECT * FROM pokemon`;
if (needsWhere) {
  query.append` WHERE 1=1`;
  if (type) query.append` AND type = ${type}`;
  if (weight) query.append` AND weight = ${weight}`;
  if (color) query.append` AND color = ${color}`;
}
const { rows } = await query.execute();
// equivalent to something like the following (depending on conditions):
const { rows } = await pool.query('SELECT * FROM pokemon WHERE 1=1 AND type = $1 AND weight = $2', [type, weight]);
```

The only breaking change this adds is that calls to `sql` now require a `.execute()` to actually send the query to the database.

### DIFF
--- a/.changeset/healthy-zebras-dress.md
+++ b/.changeset/healthy-zebras-dress.md
@@ -1,0 +1,5 @@
+---
+'@vercel/postgres': minor
+---
+
+breaking: Make `sql` a builder, requires `.execute()` to send query to db
--- a/packages/postgres/README.md
+++ b/packages/postgres/README.md
@@ -41,14 +41,22 @@ import { sql } from '@vercel/postgres';
 const id = 100;
 
 // A one-shot query
-const { rows } = await sql`SELECT * FROM users WHERE id = ${userId};`;
+const { rows } = await sql`SELECT * FROM users WHERE id = ${userId};`.execute();
 
 // Multiple queries on the same connection (improves performance)
 // warning: Do not share clients across requests and be sure to release them!
 const client = await sql.connect();
-const { rows } = await client.sql`SELECT * FROM users WHERE id = ${userId};`;
-await client.sql`UPDATE users SET status = 'satisfied' WHERE id = ${userId};`;
+const { rows } =
+  await client.sql`SELECT * FROM users WHERE id = ${userId};`.execute();
+await client.sql`UPDATE users SET status = 'satisfied' WHERE id = ${userId};`.execute();
 client.release();
+
+// Complex conditional queries
+const { rows } = await sql`SELECT * FROM `.appendUnsafeRaw('users')
+  .append` WHERE id = ${userId}`;
+// equivalent to:
+const { rows } = await sql.query('SELECT * FROM users WHERE id = $1', [userId]);
+// note that `users` was substituted as-is: `appendUnsafeRaw` does that it says on the tin. Use it carefully!
 ```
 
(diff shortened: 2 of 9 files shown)


cramforce:
Why not a new export for the builder?

elliott-with-the-longest-name-on-github:
@cramforce 
> Why not a new export for the builder?

If we weren't in beta, I'd be wholeheartedly in support of this in order to avoid a major version bump (with a plan to standardize on one API in a future major). However, we _are_ in beta, and I strongly object to unnecessarily bloating our API surface area to avoid breaking changes at this stage, _especially_ by creating another export that's almost identical to our existing one.

If we did offer another export, we'd end up with this API:

```ts
import { sql, sqlBuilder } from '@vercel/postgres';

const { rows } = await sql`SELECT * FROM users`;
const { rows } = await sqlBuilder`SELECT * FROM users`.execute();
```

With the same uncannily-similar API duplicated across the client and pool objects:

```ts
import { createPool } from '@vercel/postgres';

const pool = createPool();
const { rows } = await pool.sql`SELECT * FROM users`;
const { rows } = await pool.sqlBuilder`SELECT * FROM users`.execute();

const client = await pool.connect();
const { rows } = await client.sql`SELECT * FROM users`;
const { rows } = await client.sqlBuilder`SELECT * FROM users`.execute();
```

This package should be _simple_. Having two almost-identical-but-slightly-different ways to do the same thing is a sign that we haven't fully thought through our API.

If we bite the bullet and make this breaking change now, while we're in beta, we end up with the perfect "one way to do it all" API -- an API that's _only barely_ more complex than the one we have now:

```ts
import { sql } from '@vercel/postgres';

// simple query - only difference between old and new is `.execute()`;
const { rows } = await sql`SELECT * FROM users`.execute();

// complex query
const { type, weight, color } = filters;
const needsWhere = type || weight || color;

const query = sql`SELECT * FROM pokemon`;
if (needsWhere) {
  query.append` WHERE 1=1`;
  if (type) query.append` AND type = ${type}`;
  if (weight) query.append` AND weight = ${weight}`;
  if (color) query.append` AND color = ${color}`;
}
const { rows } = await query.execute();
```

dferber90:
I wonder if this

```js
await sql`SELECT * FROM `.appendUnsafeRaw('users').append` WHERE type = ${type};`.execute();
```

could be that

```js
import { sql, dangerouslyInsertTemplate } from '@vercel/postgres';
await sql`SELECT * FROM ${dangerouslyInsertTemplate('users')} WHERE type = ${type};`;
```

cococov:
Is there an update about this? parameterization for "ORDER BY" doesn't work either.

PR #130, opened 2023-05-08
fix: Add frozen check to `TemplateStringsArray` check

elliott-with-the-longest-name-on-github:


jridgewell (changes requested):
Babel (loose), SWC (loose), and TypeScript all transpile tagged template literals into non-frozen arrays, so this will cause bugs. While this is technically more strict and harder to misuse, it's gonna cause hisenbugs.

PR #131, opened 2023-05-09
rename kv default export

### LINKED ISSUE #107: KV: Problem importing in local mode with SvelteKit
SvelteKit (and Vite) can load environment variables from a `.env` file in development mode. SvelteKit makes those available through `$env/dynamic/private`, `$env/static/private`.

I am adding the `KV_REST_API_URL` and `KV_REST_API_TOKEN` together with my other environment variables inside the `.env` file.

When importing `@vercel/kv`, it assumes that the `KV_REST_API_URL` and `KV_REST_API_TOKEN` are present in the actual environment (through `process.env`).

In dev mode with SvelteKit, these variables are not loaded into the environment, and need to be imported. This creates an error when trying to import `@vercel/kv`. (Missing required environment variables). 

The workaround would be to `import { createClient } from '@vercel/kv'` and manually construct the proxy object, but somehow this also fails with the same environment error:

```ts
import { createClient } from '@vercel/kv'
import { KV_REST_API_TOKEN, KV_REST_API_URL } from '$env/static/private'

const kv = createClient({
	url: KV_REST_API_URL,
	token: KV_REST_API_TOKEN
})
```

correttojs:
move `kv` default export to a named export

Closes #107.

### DIFF
--- a/.changeset/witty-insects-repair.md
+++ b/.changeset/witty-insects-repair.md
@@ -1,0 +1,5 @@
+---
+'@vercel/kv': patch
+---
+
+move default export to named export
--- a/.changeset/witty-insects-repair.md
+++ b/.changeset/witty-insects-repair.md
@@ -1,5 +1,0 @@
----
-'@vercel/kv': patch
----
-
-move default export to named export
--- a/packages/kv/src/index.ts
+++ b/packages/kv/src/index.ts
@@ -80,8 +80,7 @@ export function createClient(config: RedisConfigNodejs): VercelKV {
   return new VercelKV(config);
 }
 
-// eslint-disable-next-line import/no-default-export
-export default new Proxy(
+export const kv = new Proxy(
   {},
   {
     // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
--- a/packages/kv/src/index.ts
+++ b/packages/kv/src/index.ts
@@ -80,6 +80,19 @@ export function createClient(config: RedisConfigNodejs): VercelKV {
   return new VercelKV(config);
 }
 
+// eslint-disable-next-line import/no-default-export
+export default new Proxy(
+  {},
+  {
+    // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
+    get() {
+      throw new Error(
+        '"The default export has been moved to a named export, change to import { kv }"',
+      );
+    },
+  },
+);
+
 export const kv = new Proxy(
   {},
   {
--- a/packages/kv/src/index.ts
+++ b/packages/kv/src/index.ts
@@ -85,13 +85,34 @@ export default new Proxy(
   {},
   {
     // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
-    get() {
-      throw new Error(
-        '"The default export has been moved to a named export, change to import { kv }"',
-      );
+    get(target, prop, receiver) {
+      if (prop === 'then' || prop === 'parse') {
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-return
+        return Reflect.get(target, prop, receiver);
+      }
+
+      if (!_kv) {
+        if (!process.env.KV_REST_API_URL || !process.env.KV_REST_API_TOKEN) {
+          throw new Error(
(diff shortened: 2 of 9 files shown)


test/next/package.json
```diff
@@ -6,7 +6,7 @@
     "build": "next build",
     "dev": "next dev",
     "integration-test": "playwright test",
-    "lint": "next lint",
+    "lint": "eslint --max-warnings=0 .",
```
correttojs, line 9: `next lint` caused a `lint-staging` failure

packages/kv/src/index.ts
```diff
@@ -82,6 +82,18 @@ export function createClient(config: RedisConfigNodejs): VercelKV {
 
 // eslint-disable-next-line import/no-default-export
 export default new Proxy(
+  {},
```
elliott-with-the-longest-name-on-github, line 85: Okay, I know I said we could do this, but... won't this also just make all imports fail for Vite, thereby completely defeating the point? 😂

elliott-with-the-longest-name-on-github (changes requested):

packages/kv/src/index.ts
```diff
@@ -82,6 +82,18 @@ export function createClient(config: RedisConfigNodejs): VercelKV {
...
+  {
+    // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
+    get() {
+      throw new Error(
+        '"The default export has been moved to a named export, change to import { kv }"',
+      );
+    },
```
elliott-with-the-longest-name-on-github:
```suggestion
    get(target, prop, receiver) {
      if (prop === 'then' || prop === 'parse') return Reflect.get(target, prop, receiver);
      throw new Error(
        '"The default export has been moved to a named export, change to import { kv }"',
      );
    },
```

Since this is going to be removed for V1, we could do this to prevent it from always throwing in Vite. It's the disgusting hacky solution we didn't want to implement for the default export, but since it'll go away at 1.0, I'd be comfortable with it.

elliott-with-the-longest-name-on-github: Not positive this exact thing would work (would have to test with Vite), but something like it should.

.changeset/witty-insects-repair.md
```diff
@@ -0,0 +1,5 @@
+---
+'@vercel/kv': patch
```
elliott-with-the-longest-name-on-github:
```suggestion
'@vercel/kv': minor
```

This should at least be a minor release -- obviously we're in beta, so breakings are acceptable in minors, but I would never put a breaking change in a patch release...

correttojs: oh weird, I was sure I selected the `minor` in the changeset

packages/kv/src/index.ts
correttojs: oh yeah! I wasn't sure about what you wrote, so I created a PR to discuss a practical example 😅

packages/kv/src/index.ts
correttojs: Why not add a `console.warn` to the default export, saying that
```
'"The default export will be moved to a named export after the beta, change to import { kv }"',
``` 
without throwing any error?

This way the npm version is a real `minor` (no breaking changes)

To sum up:
- `0.2.0` adds the named export, fixes Vite using the hacky solution, warns every user about the upcoming breaking change
- `1.0.0` removes the `default` export as soon as we go GA

correttojs: I was able to reproduce the `Vite` issue in a jest test, I added a test case.
I changed the `throw` to a colored `console.warn`, but indeed it may not be that visible

elliott-with-the-longest-name-on-github:
Ahhh nice, this is perfect. No breaking change until 1, a warning, and the nasty hacky fix on top. :chef-kiss:

elliott-with-the-longest-name-on-github (approved):

elliott-with-the-longest-name-on-github:
Looks like unit tests are failing though 😬

correttojs:
> Looks like unit tests are failing though 😬

oh yeah, sorry I forgot to review them!

PR #132, opened 2023-05-09
move-blob-storage

correttojs:
move blob package


test/next/src/app/vercel/blob/form-body-upload.tsx
```diff
@@ -0,0 +1,53 @@
+'use client';
+
+import type { BlobResult } from '@vercel/blob';
+import { useRef, useState } from 'react';
+
+// eslint-disable-next-line import/no-default-export
```
correttojs, line 6: @vvo I had to fix all the eslint issues in the nextjs example, including file names

elliott-with-the-longest-name-on-github (approved):
Looks good; my only complaint is that we're not actually testing anything in the integration test app. We should use Playwright to make sure this works in all of our environments.

correttojs:
> Looks good; my only complaint is that we're not actually testing anything in the integration test app. We should use Playwright to make sure this works in all of our environments.

For now I just wanted to port the existing code. As we already implemented some sort of integration tests in https://github.com/vercel/edge-functions I wanted to review the overall blob testing strategy before adding new ones

vvo:
I believe, in another PR, we would want to have a similar setup to postgres that would cover the blob package can be used in these contexts:
- api route in app directory (nodejs|edge runtimes): https://github.com/vercel/storage/tree/main/test/next/src/app/api/vercel/postgres/app
- api route in pages directory (edge|nodejs runtimes): https://github.com/vercel/storage/tree/main/test/next/src/pages/api/vercel/postgres/pages/client
- page in app directory (nodejs|edge runtimes): https://github.com/vercel/storage/tree/main/test/next/src/app/vercel/postgres/app/client
- pages in pages directory? cc @tcc-sejohnson it doesn't seem like we're covering this case as of today?

Then test all of that with https://github.com/vercel/storage/blob/main/test/next/test/%40vercel/postgres/index.test.ts. The setup from @tcc-sejohnson makes sense: you want to ensure api routes AND app router pages (server components) can use the package, is that correct @tcc-sejohnson?

To achieve that ☝️ we will have to move files around and separate pages/routes we use for development (they can stay in test/next still) from pages/routes we use for automated tests.

Thoughts @correttojs? (again, can be done in another PR).

vvo:
@correttojs Can you rebase all your commits into one, then amend its commit message so the commit comes from both of us?

Just make sure to add:

```
Co-authored-by: Vincent Voyer <vincent.voyer@vercel.com>
```

to the commit message.

test/next/src/app/vercel/blob/form-body-upload.tsx
vvo: 👍

vvo (approved):

correttojs:
> I believe, in another PR, we would want to have a similar setup to postgres that would cover the blob package can be used in these contexts:
> 
> * api route in app directory (nodejs|edge runtimes): https://github.com/vercel/storage/tree/main/test/next/src/app/api/vercel/postgres/app
> * api route in pages directory (edge|nodejs runtimes): https://github.com/vercel/storage/tree/main/test/next/src/pages/api/vercel/postgres/pages/client
> * page in app directory (nodejs|edge runtimes): https://github.com/vercel/storage/tree/main/test/next/src/app/vercel/postgres/app/client
> * pages in pages directory? cc @tcc-sejohnson it doesn't seem like we're covering this case as of today?
> 
> Then test all of that with https://github.com/vercel/storage/blob/main/test/next/test/%40vercel/postgres/index.test.ts. The setup from @tcc-sejohnson makes sense: you want to ensure api routes AND app router pages (server components) can use the package, is that correct @tcc-sejohnson?
> 
> To achieve that ☝️ we will have to move files around and separate pages/routes we use for development (they can stay in test/next still) from pages/routes we use for automated tests.
> 
> Thoughts @correttojs? (again, can be done in another PR).

Sounds good! Should we cover all the blob methods? `put`, `list`, `head`, `del`
Do we want to keep the existing blob examples somewhere?

vvo:
> Sounds good! Should we cover all the blob methods?

I don't think so, this should already be covered with unit tests. What we want to ensure: our blob package works int these contexts and can make a fetch request. Once we covered that we're good to go.

Also to add:
- an automated test that checks when uploading an image it can be used in an `<img src` tag and it works (I don't know exactly how to test that, but we want to ensure it's rendering without errors), as this is extremely common. TBH this is not really the responsibility of the blob package, but rather our infra. Or it could be covered via a synthetic test maybe?

correttojs:
> > Sounds good! Should we cover all the blob methods?
> 
> I don't think so, this should already be covered with unit tests. What we want to ensure: our blob package works int these contexts and can make a fetch request. Once we covered that we're good to go.
> 
> Also to add:
> 
> * an automated test that checks when uploading an image it can be used in an `<img src` tag and it works (I don't know exactly how to test that, but we want to ensure it's rendering without errors), as this is extremely common. TBH this is not really the responsibility of the blob package, but rather our infra. Or it could be covered via a synthetic test maybe?

Do you mean to create a Visual snapshot? I think we can do it easily in playwright calling the `toHaveScreenshot` method.

vvo:
> Do you mean to create a Visual snapshot? I think we can do it easily in playwright calling the toHaveScreenshot method.

We only have to ensure, somewhere (possibly in a synthetic test instead) that a blob url used as an image tag renders without errors, unsure the best way though for now so let's push it

vvo:
@tcc-sejohnson Do you have any information regarding 👇 

CleanShot 2023-05-10 at 11 04 56
? We're unsure where this is coming from. And if we perhaps need to be added to some project?

vvo:
We see in main @tcc-sejohnson:

CleanShot 2023-05-10 at 11 06 18

https://vercel.com/curated-tests/vercel-storage-next-integration-test-suite/HQZHmiJ9e3CqzLdECtZsH1detaEB

elliott-with-the-longest-name-on-github:
> To achieve that ☝️ we will have to move files around and separate pages/routes we use for development (they can stay in test/next still) from pages/routes we use for automated tests.

Not necessarily! Check out the routes for Postgres -- the test app is meant to be run locally so you can quickly see whether your changes are causing test failures.

elliott-with-the-longest-name-on-github:
> We see in main @tcc-sejohnson:
> 
> CleanShot 2023-05-10 at 11 06 18
> 
> https://vercel.com/curated-tests/vercel-storage-next-integration-test-suite/HQZHmiJ9e3CqzLdECtZsH1detaEB

Yeah, I don't know what's up with this. If you redeploy it _always_ works. I posted in #help-next and haven't gotten any reply.

correttojs:
> @tcc-sejohnson Do you have any information regarding 👇
> 
> CleanShot 2023-05-10 at 11 04 56 ? We're unsure where this is coming from. And if we perhaps need to be added to some project?

I see it belongs to the project https://vercel.com/vtest314-try-again/storage, which I don't think it needs to be connected

PR #140, opened 2023-05-11
add blob integration tests

correttojs:
Add blob playwright tests


.github/workflows/integration-tests-dev.yml
```diff
@@ -5,6 +5,9 @@ env:
   TURBO_TEAM: ${{ vars.TURBO_TEAM }}
   POSTGRES_URL: ${{ secrets.POSTGRES_URL }}
   POSTGRES_URL_NON_POOLING: ${{ secrets.POSTGRES_URL_NON_POOLING }}
+  BLOB_READ_WRITE_TOKEN: ${{ secrets.BLOB_READ_WRITE_TOKEN }}
```
vvo, line 11: For which store and team/project this token is? For Vercel Blob API I've used a token of one of my store but I guess we should use the same token and prefix our pathnames when necessary. Here's how it's done in Vercel Blob API: https://github.com/vercel/edge-functions/blob/26411a3960598d22ef7dbd79d3a92e57b0c2ca6e/packages/blob-storage-worker/test/index.test.ts#L17-L19

correttojs: I used https://vercel.com/curated-tests/vercel-storage-next-integration-test-suite, same as the postgres tests
It's documented in the readme, so I think it's easy to find it

vvo: > It's documented in the readme

Can you link this to me? Can't find it.

correttojs: https://github.com/vercel/storage/blob/main/test/next/README.md

test/next/src/app/vercel/blob/app/test/edge/page.tsx
```diff
@@ -0,0 +1,15 @@
...
+
+export const runtime = 'edge';
+
+export default async function AppBodyEdge(): Promise<JSX.Element> {
+  const blob = await vercelBlob.put('test-app-edge.txt', 'Hello from Pages', {
+    access: 'public',
+  });
+  return (
+    <>
+      <h1>App Router direct body upload example via an Edge Function</h1>
```
vvo: We're not receiving a body here right, we're just creating a blob immediately once the page loads, this is fine but perhaps we need to change the wording

test/next/test/@vercel/blob/index.test.ts
```diff
@@ -0,0 +1,44 @@
...
+    ].forEach((path) => {
+      test(path, async ({ request }) => {
+        const data = (await request
+          .post(`${path}?filename=test.txt`, {
+            data: 'Hello world',
+          })
+          .then((r) => r.json())) as BlobResult;
+        expect(data.contentDisposition).toBe('attachment; filename="test.txt"');
+        expect(data.contentType).toBe('text/plain');
+        expect(data.pathname).toBe('test.txt');
```
vvo: Should we add a quick test that shows, when downloaded, the blob content is the one we expect?

vvo: and a quick blob.url test, doesn't cost much to add that

test/next/test/@vercel/blob/index.test.ts
```diff
@@ -0,0 +1,44 @@
+import { test, expect } from '@playwright/test';
+import { type BlobResult } from '@vercel/blob';
+
+test.describe('@vercel/blob', () => {
+  test.describe('api', () => {
+    [
+      'vercel/blob/api/app/body/edge',
```
vvo, line 12: Do we need to test the api/ folder in both the app router and pages modes? I am unsure which ones we're covering at the moment

test/next/test/@vercel/blob/index.test.ts
```diff
@@ -0,0 +1,44 @@
+import { test, expect } from '@playwright/test';
```
vvo, line 2: I believe we need to use a prefix for all pathnames, to isolate tests. Otherwise, parallel tests from GH actions will fail (or even testing in development could fail a running GH action).
This will also allow us to cleanup after the test.

https://github.com/vercel/edge-functions/blob/26411a3960598d22ef7dbd79d3a92e57b0c2ca6e/packages/blob-storage-worker/test/index.test.ts#L17-L19

https://github.com/vercel/edge-functions/blob/26411a3960598d22ef7dbd79d3a92e57b0c2ca6e/packages/blob-storage-worker/test/index.test.ts#L42-L53

https://github.com/vercel/edge-functions/blob/26411a3960598d22ef7dbd79d3a92e57b0c2ca6e/.github/workflows/test-blob.yml#L18

.github/workflows/integration-tests-dev.yml
```diff
@@ -5,6 +5,9 @@ env:
...
+  KV_REST_API_URL: ${{ secrets.KV_REST_API_URL }}
```
vvo, line 13: Why do we need these?

vvo (changes requested):
A few comments for making changes

vvo:
@correttojs we will also need to add our tests to the live tests, correct? (when deployed on Vercel)

test/next/test/@vercel/blob/index.test.ts
correttojs: the `/pages/api` implementation is different than the `app/route` api, so I'd cover both.
I could rename `api/vercel/blob/serverless` to `api/pages/vercel/blob/serverless` to be more clear

.github/workflows/integration-tests-dev.yml
vvo: Can we be explicit and add a comment that says:

```
# This token can be found in Curated Tests > vercel-storage-next-integration-test-suite > storage > blob (which blob store is it?)
BLOB_READ_WRITE_TOKEN: ${{ secrets.BLOB_READ_WRITE_TOKEN }}
```

vvo: It's extremely common for these E2E tests tokens to be opaque after 1 year and nobody knows where do they come from.

test/next/test/@vercel/blob/index.test.ts
vvo: > the /pages/api implementation is different than the app/route api, so I'd cover both.

Yes 👍 my thinking also

test/next/test/@vercel/blob/index.test.ts
correttojs: I'm fine adding a prefix, and I agree on the cleanup.
I'm not sure why parallel tests would fail though 🤔, we always create a random URL, no?

vvo: You're right lol I missed that xD well up to you then, it's mostly for the cleanup at this point.

.github/workflows/integration-tests-dev.yml
correttojs: the `integration-tests-dev.yml` is running on a local nextjs instance, I guess it's because we want to test that the packages work on a local dev

vvo: I meant why did you had to add "KV_REST_API_URL" while this PR is for blob? Because it was missing and failing probably?

correttojs:
> @correttojs we will also need to add our tests to the live tests, correct? (when deployed on Vercel)

the live tests are exactly the same, but instead of running on a local nextjs instance they run on a preview env

.github/workflows/integration-tests-dev.yml
correttojs: not failing as we don't have tests yet, I just prepared the store and envs to add them

vvo: Got it 👍

vvo:
> the live tests are exactly the same, but instead of running on a local nextjs instance they run on a preview env

Yup, but the behavior of a deployed Next.js website on Vercel can be different (especially around uploading files: proxy!) so that's why I think we also need to have them run on Vercel.com.

test/next/src/app/vercel/blob/app/test/edge/page.tsx
correttojs: body refers to the type of fie upload...btw as we want to download the content I could show it here

vvo: > body refers to the type of fie upload

Let's call it direct string upload? I am "worried" body could be misunderstood with tests where we use request.body

> I could show it here

Indeed

correttojs:
> > the live tests are exactly the same, but instead of running on a local nextjs instance they run on a preview env
> 
> Yup, but the behavior of a deployed Next.js website on Vercel can be different (especially around uploading files: proxy!) so that's why I think we also need to have them run on Vercel.com.

I mean that all the tests defined in `test/next/test` run on `live` and `dev`

vvo (approved):
AWESOME, we will feel much more confident now :)

PR #143, opened 2023-05-11
blob-visual-test

correttojs:
This PR creates a test page that uploads and renders an image. It uses a snapshot image to ensure the image is rendered correctly.

### Reason to use Checkly
I tried to use the existing playwright setup to create this test, but I noticed the snapshot was created for the specific OS, requiring to run in a docker every time we'd need to update the snapshot.

In addition, we want to run this test on a scheduled basis, which comes out of the box with Checkly, together with the monitor integrations.

The playwright setup is still feasible in case we prefer to use the existing solution

###TODO

- [ ] add alerts
- [ ] deploy on `workflow_dispatch`?


vvo:
> every time we'd need to update the snapshot.

Is this complex to do? And if so, would we even do it at some point (once the test is written, we're good)

vvo:
Can you provide some more details around how the test really works, consider I know nothing about checkly. What does the test does once it took a screenshot? And how does that ensures an image can be rendered?

vvo:
Can you provide some more details around how the test really works, consider I know nothing about checkly. What does the test does once it took a screenshot? And how does that ensures an image can be rendered?

correttojs:
> Can you provide some more details around how the test really works, consider I know nothing about checkly. What does the test does once it took a screenshot? And how does that ensures an image can be rendered?

I notice that the test works only in case we call `toHaveScreenshot` against an existing screenshot, which checkly doesn't take care of storing 😞. This makes it harder to update the checkly snapshot anyway (similar to the docker solution)

I could go with the playwright solution then, the only issue is to create alerts in case of failures. Do you have any solution in mind, without the need of creating a custom GH action?

vvo:
> I could go with the playwright solution then, the only issue is to create alerts in case of failures.

Because you would configure the github action to run on a schedule? Otherwise I am not getting it, since a build would trigger the failure already.

correttojs:
> > I could go with the playwright solution then, the only issue is to create alerts in case of failures.
> 
> Because you would configure the github action to run on a schedule? Otherwise I am not getting it, since a build would trigger the failure already.

The idea is to run this test on a schedule, right? In this case we need to integrate an alert when it fails

vvo:
> The idea is to run this test on a schedule, right? In this case we need to integrate an alert when it fails

Could we run the GitHub action via a synthetic test?

vvo:
@correttojs Could we use https://docs.datadoghq.com/synthetics/browser_tests/actions maybe? If the main branch of this repository is always deployed on the same url, we can create a synthetic test that ensures the very basic functionality always works (upload file, test that it renders, done).

vvo:
> The idea is to run this test on a schedule, right?

First the idea is to run this test on PRs, then if we can also run it on a schedule it's great. But otherwise we can just defer to Datadog synthetic tests for the schedule part.

vvo:
I guess the tricky part is to find a reliable way to express "test that this image rendered" which I am not sure what's the best way to do with either playwright/cypress/datadog synthetic tests

correttojs:
> I guess the tricky part is to find a reliable way to express "test that this image rendered" which I am not sure what's the best way to do with either playwright/cypress/datadog synthetic tests

I think the best way to achieve that is using (visual) snapshot tests, which are available in playwright through the `toHaveScreenshot()` [api](https://playwright.dev/docs/test-snapshots). 
Checkly is a wrapper (with its own runners) on top of Playwright, so if it works it would be the simplest solution as the it offers monitors out of the box.
I already contacted the Checkly support: if this option is not feasible, I'd go with a GH action, sending failure results to slack/PD

The Datadog browser tests are still quite limited, it doesn't seem they offer a snapshot solution (or anything similar to assert that the image is loaded)

vvo:
Maybe https://github.com/ravsamhq/notify-slack-action could help to send a notification when there's a failure

vvo:
Or use https://www.datadoghq.com/blog/datadog-github-actions-ci-visibility/?

correttojs:
Checkly confirmed they don't support Visual regression tests, though it's on their roadmap.

They also provided a nice idea to assert the snapshot file size: you cannot compare the current and the prev snapshot, but you can get the size once and store its value. Not 100% accurate but could work for our case

To recap:
- Option 1: Playwright calls `toHaveScreenshot` and stores it in a local file (using a docker so the OS and the snapshot filename are the same as the CI). Use  https://www.datadoghq.com/blog/datadog-github-actions-ci-visibility/ to monitor failures
- Option 2: Checkly calls `page.screenshot()` and compares the image size against an hardcoded value. Monitors are setup using Checkly

vvo:
Which option would you prefer? I am fine with both as long as we get an alert if something fails.

correttojs:
> Which option would you prefer? I am fine with both as long as we get an alert if something fails.

I'll give a try to the datadog CI integration...having a real visual regression is nicer, and using a single testing framework would reduce the noise

vvo:
@correttojs Do you still need another review on this PR?

correttojs:
> @correttojs Do you still need another review on this PR?

I will close it as soon as https://github.com/vercel/storage/pull/149 (playwright version) will work

PR #147, opened 2023-05-14
feat: Change `KV_URL` to `KV_REST_API_URL`

LWJerri:
We can't use `KV_URL` because it returns a Redis connection string, so we need to use `KV_REST_API_URL`, which gives the https connection string.


elliott-with-the-longest-name-on-github:
I'm not sure why this needs to change -- it's currently working on every platform I've tried it on. Can you please file an issue if something is broken?

correttojs (approved):

correttojs:
You're right @LWJerri, thanks for reporting the issue!

PR #149, opened 2023-05-15
blob-visual-test-playwright

correttojs:
Add blob visual regression test

- the new schedule pipeline run [here](https://github.com/vercel/storage/actions/runs/4979144604/jobs/8910378027?pr=149)
- add playwright visual test https://playwright.dev/docs/test-snapshots
- add datadog action to update the `blob.storage.load.image` service-check
- created a Datadog monitor linked to the new service-check https://app.datadoghq.com/monitors/119037464

Note: the tests are running on every PR (on `integration-tests-live` and `integration-tests-dev`) and on a schedule basis. Locally they run together with the playwright test suite. The snapshots references are available for `darwin` and `linux`


package.json
```diff
@@ -25,9 +25,6 @@
   "lint-staged": {
     "*": [
       "prettier --ignore-unknown --write"
-    ],
```
correttojs, line 28: I removed this pre-commit hook as it's really slow (as it requires to `build` everything). The CI runs a `lint` check anyway, so it's not mandatory

.github/workflows/visual-tests-schedule.yml
```diff
@@ -0,0 +1,83 @@
...
+          key: ${{ runner.os }}-playwright-${{ hashFiles('**/pnpm-lock.yaml') }}
+
+      - name: Install dependencies
+        run: pnpm install
+
+      # This will use the cache for browsers. But it will reinstall system deps, which can't be cached
+      # https://github.com/microsoft/playwright/issues/20603#issuecomment-1416243810
+      - run: pnpm dlx playwright install --with-deps
+
+      - name: Run Playwright tests on https://storage-rho.vercel.app/
```
vvo, line 52: What is storage-rho?

correttojs: the production link of https://vercel.com/curated-tests/vercel-storage-next-integration-test-suite

.github/workflows/visual-tests-schedule.yml
```diff
@@ -0,0 +1,83 @@
...
+        run: pnpm integration-test
```
vvo, line 53: To confirm: the visual regression test is part of the main integration-test suite. And now we will run the main integration-test suite on a schedule.

vvo: And we will only run the visual test part 👍

correttojs: right, this scheduled workflow defines the `PLAYWRIGHT_TEST_MATCH` which executes only the visual regression tests

.github/workflows/visual-tests-schedule.yml
```diff
@@ -0,0 +1,83 @@
+name: Playwright Visual Tests (Scheduled)
```
vvo, line 1: Should we run tests only in Chrome Desktop on Linux? I don't think we need to run tests in multiple platforms. If it works on Chrome Linux (or any other easy to run in CI platform), then it will work everywhere.

.github/workflows/visual-tests-schedule.yml
vvo: Does rho means anything in particular or just random characters?

vvo (approved):
Aside from advice to run only in one browser to reduce flakiness, all good for me.

.github/workflows/visual-tests-schedule.yml
correttojs: we could run them on a single browser, I opted to minimize the changes needed in the playwright config, and these tests seem quite stable, so I wouldn't mind running them on the 3 defined browsers

vvo: if it's fast enough and stable, all good

.github/workflows/visual-tests-schedule.yml
correttojs: That's what vercel defined as default production link. We could create an alias eventually

correttojs:
> Aside from advice to run only in one browser to reduce flakiness, all good for me.

Ok I run only on Chrome within the schedule. Note that locally and in the PRs all the integration tests (including VRT) will run in all the browsers, so we need to keep the related screenshots

vvo:
With the merge of #163 this should now work 👍

Issue #119, opened 2023-05-05
Error: @vercel/edge-config: Network error 


Metarock:
### Environment information
```   Operating System:
      Platform: linux
      Arch: x64
      Version: #202303130630~1681329778~22.04~d824cd4 SMP PREEMPT_DYNAMIC Wed A
    Binaries:
      Node: 18.15.0
      npm: 9.5.0
      Yarn: 1.22.19
      pnpm: 7.29.0
    Relevant packages:
      next: 13.4.0
      eslint-config-next: 13.4.0
      react: 18.2.0
      react-dom: 18.2.0
```

### To reproduce

1. Use the latest Nextjs version app directory
2. use a `generateStaticParams` on a server component page `app/[id]/yourpage.tsx` using a edge-config SDK function, in my case the `get` call 

### Describe the bug
When building a static page in Nextjs with `generateStaticParams` that uses the edge-config SDK it spits out the following error
```error - node_modules/@vercel/edge-config/dist/index.node.js (1:1052) @ eval
error - Error: @vercel/edge-config: Network error
    at eval (webpack-internal:///(sc_server)/./node_modules/@vercel/edge-config/dist/index.node.js:46:23)
    at async Layout (webpack-internal:///(sc_server)/./app/sites/[site]/layout.tsx:25:20) {
  digest: undefined
}
null
```

### Expected behavior
I expect the edge-config SDK to not spit out an error.  For now as a work around  I have opted out from using the SDK client and go with the Vercel REST API for now as it seems to be working as expected `https://edge-config.vercel.com/your_edge_config_id_here/item/example_key_1?token=your_edge_config_read_access_token_here`

dferber90:
We identified an issue in Next.js which is likely causing this. Will update here once a fix is out

dferber90:
While the issue currently remains inside of Next.js, we temporarily added a workaround for this in https://github.com/vercel/storage/pull/157

It's available as `@vercel/edge-config@0.1.11`

jeremieP:
Oh thanks @dferber90 ! 

Will test it this afternoon on my project, and make you a feedback if necessary.

UPDATE : Worked perfectly on my project ! Thanks again !

danielbichuetti:
We have encountered this issue while utilizing Nuxt. 100% of the time.

dferber90:
@danielbichuetti You've already pointed this out here https://github.com/vercel/storage/issues/95#issuecomment-1913282484.

[...]

PR #152, opened 2023-05-18
add reproduction and workaround

dferber90:
This is the beginning of a workaround for #119, but
- issue does not reproduce in our test/next example as root layout declares `export const dynamic = 'force-dynamic';` so I had to disable it temporarily
- the solution we'll add here is going to be a workaround only, the real fix will need to happen in next.js later
  - once next.js fixes it, we should remove our workaround


test/next/src/app/layout.tsx
```diff
@@ -2,8 +2,6 @@ import { Inter } from 'next/font/google';
 
 const inter = Inter({ subsets: ['latin'] });
 
-export const dynamic = 'force-dynamic';
```
dferber90, line 5: @tcc-sejohnson I had to remove this in order to reproduce issue-119. I created and committed a reproduction for 119. I was only able to reproduce by not having force-dynamic declared globally.

I instead added it to the selected pages manually.

dferber90:
Going to open a new PR due to the amount of conflicts.

PR #154, opened 2023-05-18
unify edge and node implementations

dferber90:
Our `@vercel/edge-config` needs to work well in both `edge-light` and `nodejs` runtimes.

The root hurdle is that we need to import `fs` for `nodejs` but this is not allowed in `edge-light`. 

Until now we've solved this by using two separate exports, one for `nodejs` and one for `edge-light`. This allowed us to have two separate entry points. But having two separate entry points means we end up duplicating a lot of code.

A much better approach is to have a single entry point and to then be able to conditionally branch within.

I tried to do this when first creating the package but it was not possible due to [import maps](https://github.com/WICG/import-maps) not being supported by typescript.

The big realization now was that we can fake it by having a second separate module.

This allows us to
- remove deduplication of `index.edge.ts` and `index.node.ts`
- finally have JSDoc
- remove around 170 lines of duplicated code


packages/edge-config/src/index.node.test.ts
AAorris, line 1: Should we rename `index.node.test.ts` to `index.test.ts` to stay in line with the source file?

packages/edge-config/src/index.common.test.ts
```diff
@@ -27,168 +17,164 @@ describe('test conditions', () => {
...
+      fetchMock.mockResponseOnce('', {
+        status: 304,
+        headers: { ETag: 'a' },
+      });
+
+      // second call should reuse response
+
+      await expect(edgeConfig.get('foo')).resolves.toEqual('bar');
+
+      expect(fetchMock).toHaveBeenCalledTimes(2);
```
AAorris, line 160: Should we assert that readFile is not called as we do in the node tests?

AAorris (approved):

packages/edge-config/src/index.common.test.ts
dferber90: If it was called in edge it would throw, so I think we're good.

We can follow up in another PR if we want to clean up more 👍

packages/edge-config/src/index.node.test.ts
dferber90: There's probably a lot of duplication in the tests that we can get rid of 👍 

For this PR I wanted to mostly not touch the tests so we can have confidence they still test the same thing. We can dedupe in a follow up after this PR.

For now I'd like to keep our structure of
- `index.common.test.ts`
- `index.edge.test.ts`
- `index.node.test.ts`

PR #157, opened 2023-05-19
work around nextjs issue affecting async page components in app router on the nodejs runtime

dferber90:
This a workaround for #119, but
- issue does not reproduce in our test/next example as root layout declares `export const dynamic = 'force-dynamic';` so I had to disable it temporarily
- the solution we'll add here is going to be a workaround only, the real fix will need to happen in next.js later
  - once next.js fixes it, we should remove our workaround

Based on #152


test/next/src/app/layout.tsx
```diff
@@ -2,8 +2,6 @@ import { Inter } from 'next/font/google';
 
 const inter = Inter({ subsets: ['latin'] });
 
-export const dynamic = 'force-dynamic';
```
dferber90, line 5: Added it to the individual sub-pages instead so we don't force `force-dynamic` for `edge-config`, otherwise it prevents the reproduction of a specific bug there.

AndyBitz (approved):

PR #169, opened 2023-05-23
Fix issue with Rollup and Vite

### LINKED ISSUE #168: @vercel/kv: Missing required environment variables KV_REST_API_URL and KV_REST_API_TOKEN
Vercel KV works when running locally, but not when running from an edge function.

Here's my code:
```typescript
import { kv } from "@vercel/kv";

//the following is inside an Edge Function:
console.log("keys", await kv.keys("*"));
```

# My local setup (which works)
Ran the following commands:

```bash
vercel env pull .env.ssr.local
vite --mode ssr
```

When I invoke my function, it works just fine and sets the content in the KV store. I can verify this by going to the CLI in the Dashboard and checking that the key is there.

# My Vercel Edge Function setup (which doesn't work)
Pushed to my repo. 

Had Vercel deploy it. 

Verified in the build logs that both the function was deployed and the static assets:
image

Verified that my database is attached to the project in the dashboard:
image

After triggering the edge function, it returns an error 500. Viewing the logs yields the following error:

```
Error: @vercel/kv: Missing required environment variables KV_REST_API_URL and KV_REST_API_TOKEN
    at (@qwik-city-plan.js:27:78546)
    at (@qwik-city-plan.js:27:85712)
    at (assets/index.qwik-cc4e9f9f.js:9:5088)
    at (assets/index.qwik-cc4e9f9f.js:13:2233)
    at (assets/index.qwik-cc4e9f9f.js:7:1389)
    at (assets/index.qwik-cc4e9f9f.js:13:2161)
    at (assets/index.qwik-cc4e9f9f.js:13:2394)
    at (entry.vercel-edge.js:16:441)
    at (entry.vercel-edge.js:20:3149)
    at (entry.vercel-edge.js:20:830)
```

ffMathy:
Fixes #168.

It seems many tools like Vite and Rollup will remove `process.env` calls, unless the string version is used instead.

### DIFF
--- a/.changeset/metal-rabbits-work.md
+++ b/.changeset/metal-rabbits-work.md
@@ -1,0 +1,5 @@
+---
+"@vercel/kv": patch
+---
+
+Fix issue with Rollup and Vite
--- a/packages/kv/src/index.ts
+++ b/packages/kv/src/index.ts
@@ -120,15 +120,15 @@ export const kv = new Proxy(
     // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
     get(target, prop) {
       if (!_kv) {
-        if (!process.env.KV_REST_API_URL || !process.env.KV_REST_API_TOKEN) {
+        if (!process.env['KV_REST_API_URL'] || !process.env['KV_REST_API_TOKEN']) {
           throw new Error(
             '@vercel/kv: Missing required environment variables KV_REST_API_URL and KV_REST_API_TOKEN',
           );
         }
 
         _kv = createClient({
-          url: process.env.KV_REST_API_URL,
-          token: process.env.KV_REST_API_TOKEN,
+          url: process.env['KV_REST_API_URL'],
+          token: process.env['KV_REST_API_TOKEN'],
         });
       }
 


vvo:
It seems there's a workaround in our documentation, let's see first if that works: https://github.com/vercel/storage/tree/main/packages/kv#a-note-for-vite-users

As this current workaround in your PR might just stop working for vite users at another point in time? Also good read: https://vitejs.dev/guide/env-and-mode.html

ffMathy:
No doubt that it's a hack, but it can only get better, even if it's temporary. I tried following both of those articles, and none of them worked for me.

elliott-with-the-longest-name-on-github:
@ffMathy Are you using a very old version of Vite? It used to do this but hasn't for a very long time.

ffMathy:
Nope.

Brand new Qwik project created 3 weeks ago.

It's "poodle.lol" on GitHub.

ffMathy:
Remember that Rollup does it too. And Rollup is part of Qwik.

ffMathy:
Honestly, I don't understand the current reasoning of not merging this.

Help me understand. We have a couple of possible outcomes:

1. Rollup or Vite or whatever never changes their logic, and the hack keeps working forever. 
2. They change their logic. Then it falls back to where we were.
3. We accidentally change the process.env calls back to how they were. To prevent this, we can even extract the retrieval of the env variables to a separate function, and then link to this issue in that function to help ensure it stays that way.

Or am I missing something? 

The "hack" is not even doing anything nasty from a point of JavaScript's perspective.

So we are winning a lot, and currently losing nothing. Why not take those odds?

vvo:
@ffMathy Thanks for the feedback, we'd definitely want to fix this issue yes. One reason I can suggest as to why we haven't yet merged this is that we haven't found a documentation about vite, rollup or quik that says "If you want your npm package to be compatible with Vite, Rollup or Quik then you should use process.env['VAR_NAME'] and not process.env.VAR_NAME".

We trust you this is working for you yes. But we'd like to better understand the root cause for "all of that" (it's a small change, correct).

What I can suggest to unblock you immediately until we can get a confirmation that this is required in all cases, is to use https://github.com/ds300/patch-package (I've used it multiple times) so you can apply your patch to your deployment and keep on working.

Thanks!

vvo:
Relevant: https://github.com/BuilderIO/qwik/issues/3962

vvo:
@ffMathy So it looks like this is a Quik bug mostly (and Vite/Rollup recent not-part-of-quik versions would be working?)

cc @Schniz would the fix from @ffMathy be safe here?

Also: https://github.com/BuilderIO/qwik/pull/4361
Note: you experienced this bug in an edge function only

vvo:
See https://github.com/vercel/storage/issues/168#issuecomment-1787317149

PR #173, opened 2023-05-30
[edge-config] Add env and SDK version header to requests

AndyBitz:
- Adds `x-edge-config-vercel-env` as a header to requests to provide a way to differentiate between development and other origins
 - Adds `x-edge-config-sdk` as a header to requests to see the version of the SDK that was used

### DIFF
--- a/packages/edge-config/src/index.common.test.ts
+++ b/packages/edge-config/src/index.common.test.ts
@@ -80,7 +80,7 @@ describe('when running without lambda layer or via edge function', () => {
           {
             headers: new Headers({
               Authorization: 'Bearer token-2',
-              'x-node-env': 'test',
+              'x-edge-config-vercel-env': 'test',
             }),
             cache: 'no-store',
           },
@@ -103,7 +103,7 @@ describe('when running without lambda layer or via edge function', () => {
             method: 'HEAD',
             headers: new Headers({
               Authorization: 'Bearer token-2',
-              'x-node-env': 'test',
+              'x-edge-config-vercel-env': 'test',
             }),
             cache: 'no-store',
           },
@@ -125,7 +125,7 @@ describe('when running without lambda layer or via edge function', () => {
           {
             headers: new Headers({
               Authorization: 'Bearer token-2',
-              'x-node-env': 'test',
+              'x-edge-config-vercel-env': 'test',
             }),
             cache: 'no-store',
           },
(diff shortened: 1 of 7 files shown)


packages/edge-config/src/index.common.test.ts
```diff
@@ -78,7 +78,10 @@ describe('when running without lambda layer or via edge function', () => {
         expect(fetchMock).toHaveBeenCalledWith(
           `${modifiedBaseUrl}/item/foo?version=1`,
           {
-            headers: new Headers({ Authorization: 'Bearer token-2' }),
+            headers: new Headers({
+              Authorization: 'Bearer token-2',
+              'x-node-env': 'test',
```
dferber90: Is x-node-env something we use anywhere else at vercel? Otherwise I’d maybe name this x-edge-config-node-env

packages/edge-config/src/index.ts
```diff
@@ -164,7 +164,12 @@ export function createClient(
 
   const baseUrl = connection.baseUrl;
   const version = connection.version; // version of the edge config read access api we talk to
-  const headers = { Authorization: `Bearer ${connection.token}` };
+  const headers: Record<string, string> = {
+    Authorization: `Bearer ${connection.token}`,
+  };
+
+  if (typeof process !== 'undefined' && process.env.NODE_ENV)
```
dferber90: It’s a bit weird that we’ll be sending NODE_ENV even from edge functions and middleware since it’s not node

dferber90: Im wondering if VERCEL_ENV would be a better candidate since NODE_ENV is often set to production even when working locally

dferber90: Should we also add the edge config SDKs version to the request headers here?

packages/edge-config/src/index.common.test.ts
AndyBitz: Not that I'm aware of, happy to change

packages/edge-config/src/index.ts
AndyBitz: For consistency. I guess, ideally, the SDK wouldn't even know in which type of compute it runs.

AndyBitz: Good point, makes sense. NODE_ENV is probably still more common if someone wouldn't use `vc env pull` etc. but VERCEL_ENV should probably provide the same insights.

dferber90: Since the refactoring which merged the entry points it isn’t as explicitly aware of it anymore as it was before. My comment was mainly about how we collectively ended up with having NODE_ENV in a runtime which isn’t actually node. No particular point, just found it a bit humorous

AndyBitz: Ohh, haven't even thought about that, true

AndyBitz: Added

dferber90 (approved):

PR #179, opened 2023-06-06
[blob] implement client upload

correttojs:
### implement client upload

- implement `generateClientTokenFromReadWriteToken`
- add client upload test page for serverless and edge


packages/blob/client/package.json
```diff
@@ -0,0 +1,41 @@
+{
```
dferber90: This doesn’t need so much detail afaik. That info will get used from the main package for npm. You can get away with just having the main/exports, type and types fields

packages/blob/package.json
```diff
@@ -21,6 +21,16 @@
         "node": "./dist/index.cjs",
         "default": "./dist/index.cjs"
       }
+    },
+    "./client": {
```
dferber90: This doesn’t need to be here at all iirc

vvo:
Once ready, let's ship this as experimental/next tag on npm so we can give a try to the published built package to ensure endpoints and types are correctly working.

packages/blob/src/index.ts
```diff
@@ -215,8 +223,42 @@ export async function list(
...
+  onUploadCompletedUrl: string;
+}
+
+export function generateUploadTokenFromReadWriteToken({
+  pathname,
+  onUploadCompletedUrl,
+  token,
+}: GenerateUploadUrlOptions): string {
+  const timestamp = new Date();
+  timestamp.setSeconds(timestamp.getSeconds() + 30);
```
vvo, line 254: Let's have a default timestamp that is longer than 30s. Unless you think this would introduce errors? I think we can have something like 2 minutes, and this should not be an issue.

vvo:
@correttojs Should we allow providing custom metadata at token creation time? When using S3-presigned URLs you can add metadata. Here's why I think we should think about a way to do so: As a customer using this feature for avatar uploads (example), when onUploadCompletedCallbackUrl triggers, then it will contain the url of the avatar.

Now as a customer, I'd like to store this url in my database to update the avatar of the right user. To do so I can either:
- parse the url if it contains the user id (perhaps it doesn't)
- or use metadata from the onUploadCompletedCallbackUrl (= the metadata needs to be set at token creation time).

I propose we allow this:
```
generateUploadTokenFromReadWriteToken({
  pathname: 'foo.txt',
  onUploadCompletedUrl: 'https://example.com/',
  metadata: 'userId=314' // any string is ok, this can be used by customers and it would contain comma separated values, json, whatever we don't care
  token: 'TEST_TOKEN',
});
```

Thoughts?

packages/blob/src/index.ts
correttojs: sure 2 mins is fine as well. I considered that the token needs to be valid for the time the client makes a request to CF and CF calls the vercel validation endpoint, which should be pretty low. Increasing the token time would increase the risk of multiple requests, though these would be quite useless as the user can upload only a file with the same pathname

vvo: I am thinking about a page that contains a `<form>` with many fields, some of them being file uploads (going directly to Cloudflare with our technique). Probably the token would be generated once at page load and that's it. In this case we should probably have a default token duration that is long enough (and even 2 minutes may not be sufficient).

That, or we should document how to write a useClientUploadToken that relies on SWR to revalidate the token when time has passed (this may be the best option overall).

packages/blob/package.json
```diff
@@ -33,7 +44,7 @@
     "dist"
   ],
   "scripts": {
-    "build": "tsup && cp src/undici-browser.js dist/undici-browser.js",
+    "build": "tsup  && cp src/undici-browser.js dist/undici-browser.js ",
```
vvo: are the extra spaces needed?

vvo:
Once we make this right, let's make sure we update the README with an example that uses this feature. With a similar code than the current README example.

packages/blob/src/index.ts
correttojs: even if you have multiple files in the same form, I'd always couple the token generation to the `blob.put` request. Otherwise you cannot say for how much time a token should be valid

packages/blob/src/client.ts
```diff
@@ -0,0 +1,72 @@
...
+  if (!body) {
+    throw new BlobError('body is required');
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+  if (!options || options.access !== 'public') {
+    throw new BlobError('access must be "public"');
+  }
+
+  if (!options.clientToken) {
```
vvo: Can we keep "token"? Since we're already importing from client I do not think we need to change the option name here

packages/blob/src/index.ts
vvo: Oh right, multiple files I did not even think about that (soon enough we will need prefix: option for generate blob token to solve that easily, I'll make a note of it)

vvo: > even if you have multiple files in the same form, I'd always couple the token generation to the blob.put request. Otherwise you cannot say for how much time a token should be valid

You're right, even more since the token will be used once, let's make sure our default example in README does a submitHandler + generateToken + uploadFile then (correct?).

correttojs: is it fine if we keep 30s then?

vvo: Yep I think so for now it's fine

packages/blob/src/index.test.ts
```diff
@@ -481,4 +487,37 @@ describe('blob client', () => {
...
+    afterEach(() => {
+      jest.runOnlyPendingTimers();
+      jest.useRealTimers();
+    });
+
+    beforeEach(() => {
+      jest.useFakeTimers().setSystemTime(new Date('2023-01-01'));
+    });
+    test('should generate an upload token with the correct payload', () => {
+      const uploadToken = generateUploadTokenFromReadWriteToken({
```
vvo: So we do not forget, this should be `generateClientTokenFromReadWriteToken` right?

correttojs:
> @correttojs Should we allow providing custom metadata at token creation time? When using S3-presigned URLs you can add metadata. Here's why I think we should think about a way to do so: As a customer using this feature for avatar uploads (example), when onUploadCompletedCallbackUrl triggers, then it will contain the url of the avatar.
> 
> Now as a customer, I'd like to store this url in my database to update the avatar of the right user. To do so I can either:
> 
> * parse the url if it contains the user id (perhaps it doesn't)
> * or use metadata from the onUploadCompletedCallbackUrl (= the metadata needs to be set at token creation time).
> 
> I propose we allow this:
> 
> ```
> generateUploadTokenFromReadWriteToken({
>   pathname: 'foo.txt',
>   onUploadCompletedUrl: 'https://example.com/',
>   metadata: 'userId=314' // any string is ok, this can be used by customers and it would contain comma separated values, json, whatever we don't care
>   token: 'TEST_TOKEN',
> });
> ```
> 
> Thoughts?

@vvo  added `onUploadCompletedCallbackUrlArgs`, feel free to suggest a better name

packages/blob/src/client.ts
correttojs: throwing in case the token doesn't start with `vercel_blob_client`

packages/blob/package.json
correttojs: I was playing with the tsup config, spaces removed

.github/workflows/integration-tests-dev.yml
```diff
@@ -47,7 +47,7 @@ jobs:
 
       # This will use the cache for browsers. But it will reinstall system deps, which can't be cached
       # https://github.com/microsoft/playwright/issues/20603#issuecomment-1416243810
-      - run: pnpm dlx playwright install --with-deps
+      - run: pnpm dlx playwright@1.34.0 install --with-deps
```
correttojs: this fixes the playwright dependency mismatch

vvo: I propose we use the less-pretty fix of forcing which binary to use in https://github.com/vercel/storage/pull/183/files, so versioning only happens at the package.json level

correttojs: I'll rebase it once merged

packages/blob/src/index.ts
```diff
@@ -215,8 +211,60 @@ export async function list(
...
 
-function getApiUrl(): string {
-  return process.env.VERCEL_BLOB_API_URL || 'https://blob.vercel-storage.com';
+export interface GenerateClientTokenOptions extends BlobCommandOptions {
+  pathname: string;
+  onUploadCompletedUrl: string;
+  onUploadCompletedCallbackUrlArgs?: Record<string, string>;
+}
+
+export function generateClientTokenFromReadWriteToken({
```
correttojs: @vvo despite the additional network request, I think moving this function to the vercel api (or the worker eventually) would allow us to update the client token format without breaking changes on the `npm` module. wdyt?

packages/blob/package.json
```diff
@@ -11,26 +11,16 @@
   "license": "Apache-2.0",
   "sideEffects": false,
   "type": "module",
-  "exports": {
```
vvo, line 14: Are you 100% sure this will generate a package that will work when imported in:
- when used in a regular node.js script
- as a nextjs application dependency
- an edge function in nextjs
- a browser

For some reason, I remember having spent a good amount of time to find we had to put these lines. If you manage to check it's working without them in the various contexts fine 👍

packages/blob/src/index.ts
```diff
@@ -215,8 +211,60 @@ export async function list(
   };
 }
 
-function getApiUrl(): string {
-  return process.env.VERCEL_BLOB_API_URL || 'https://blob.vercel-storage.com';
+export interface GenerateClientTokenOptions extends BlobCommandOptions {
+  pathname: string;
+  onUploadCompletedUrl: string;
```
vvo: This should be `onUploadCompletedCallbackUrl` right?

packages/blob/src/index.ts
```diff
@@ -215,8 +211,60 @@ export async function list(
...
+  onUploadCompletedCallbackUrlArgs?: Record<string, string>;
```
vvo: Perhaps both parameters would be easier to understand this way:

```
generateClientTokenFromReadWriteToken({
  onUploadCompleted: {
    callbackUrl: '...',
    metadata: {...}
  }
})
```

thoughts?

Since metadata can't be used without the callback. And it makes naming easier.

packages/blob/src/index.ts
vvo: > would allow us to update the client token format without breaking changes on the npm module.

What kind of changes can you foresee at the moment? The fact that even AWS did it this way makes me think this is a correct way, even if you ever have to implement changes.

packages/blob/package.json
correttojs: I created e2e tests for edge, node and browser. I also run the script 
```
const blob = require('@vercel/blob');

blob
  .list({
    token: '...',
  })
  .then((r) => console.log(r));

```
inside `test/next`

packages/blob/src/index.ts
```diff
@@ -215,8 +211,60 @@ export async function list(
...
+  return `vercel_blob_client_${storeId}_${Buffer.from(
+    `${securedKey}.${payload}`,
+  ).toString('base64')}`;
+}
+
+type DecodedClientTokenPayload = Omit<GenerateClientTokenOptions, 'token'> & {
+  validUntil: number;
+};
+
+export function decodeClientToken(
```
vvo: This gets exported at the package level, it should not be exported I believe?
If we need this for tests, we can have it in a different file instead and use it in tests only.

Also you could name it "getPayloadFromClientToken" since it doesn't really decode it but give you the raw payload part.

packages/blob/src/index.ts
```diff
@@ -215,8 +211,58 @@ export async function list(
   };
 }
 
-function getApiUrl(): string {
-  return process.env.VERCEL_BLOB_API_URL || 'https://blob.vercel-storage.com';
+export interface GenerateClientTokenOptions extends BlobCommandOptions {
+  pathname: string;
+  onUploadCompleted?: {
+    callbackUrl: string;
+    metadata?: Record<string, string>;
```
vvo: @correttojs metadata must be a string. We have no idea how folks will use it. Some will put objects in it, some simple strings, some query strings, ... Let's accept any string and delegate JSON.parse/JSON.stringify to the caller/receiver instead.

packages/blob/src/index.ts
correttojs: I would export it and use on the vercel api, but I think it would be a useful helper for the user

vvo: > but I think it would be a useful helper for the user

Curious: Any cases you can think of this would be useful for users?

correttojs: if you think about the jwt, the client can always access the payload and decode if needed

packages/blob/src/index.ts
```diff
@@ -215,8 +211,58 @@ export async function list(
...
+    JSON.stringify({
+      ...args,
+      validUntil: timestamp.getTime(),
+    }),
+  ).toString('base64');
+
+  const [, , , storeId = null] = blobToken.split('_');
+
+  if (!storeId) {
+    throw new Error('Invalid BLOB_READ_WRITE_TOKEN');
```
vvo: This is not the correct error if you pass token: in the params of this method. Also if we check the storeId perhaps we should do so before const payload which comes before at the moment (the optimizer will do this, this is more for readability)?

packages/blob/src/index.ts
correttojs: why not `any` then?

vvo: any would mean function/regex/dates are valid types wise, while when receiving the metadata, the format will no more work (a date would no more be a date unless we reparse the metadata..).

`string` is a low-level option that is flexible for all users without forcing them to send key/value when they don't need to.

correttojs: I'd still prefer a generic `TMetadata = any`, but I don't have a strong preference so `string` is also fine

packages/blob/package.json
vvo: 👍

correttojs: I could put it back as the change is irrelevant to the solution. anyway, it doesn't seem to be needed

vvo: @correttojs if these lines are not needed for the module to work in various situations then yes we can remove them 👍

packages/blob/README.md
```diff
@@ -202,6 +218,104 @@ export async function POST(request: Request) {
...
+// /app/api/upload-token/route.ts
+
+import {
+  generateClientTokenFromReadWriteToken,
+  type GenerateClientTokenOptions,
+} from '@vercel/blob';
+import { NextResponse } from 'next/server';
+
+export async function POST(request: Request): Promise<NextResponse> {
+  const body = (await request.json()) as GenerateClientTokenOptions;
```
vvo: I don't think we should reuse this type at this point.
Because metadata and callbackUrl can't be set by the browser.

We better just accept pathname

```suggestion
  // On a real website, this route would be protected by authentication, see: https://nextjs.org/docs/pages/building-your-application/routing/authenticating
  // Here, we accept the `pathname` from the browser, but in some situations, you may even craft the pathname
  // based on the authentication result
  const body = (await request.json()) as { pathname: string };
```

packages/blob/README.md
```diff
@@ -202,6 +218,104 @@ export async function POST(request: Request) {
...
+```
+
+```ts
+// /app/api/upload-completed/route.ts
+
+import { type BlobUploadCompletedEvent } from '@vercel/blob';
+import { NextResponse } from 'next/server';
+
+export async function POST(request: Request): Promise<NextResponse> {
+  const body = (await request.json()) as BlobUploadCompletedEvent;
```
vvo, line 327: TODO: check signature

correttojs: https://linear.app/vercel/issue/STO-656/implement-webhook-signature-verification-helper

packages/blob/README.md
```diff
@@ -128,6 +129,23 @@ async function list(options?: {
 }> {}
 ```
 
+### generateClientTokenFromReadWriteToken(options)
+
+Generates a single-use token which can be used from within the client.
```
```suggestion
Generates a single-use token that can be used from within the client. This is useful when [uploading directly from browsers](#uploading-directly-from-browsers) to circumvent the 4MB limitation of going through a Vercel-hosted route.
```

Not sure about the markdown link 👍

packages/blob/README.md
```diff
@@ -128,6 +129,23 @@ async function list(options?: {
...
+  token?: string;
+  pathname?: string;
+  onUploadCompleted?: {
+    callbackUrl: string;
+    metadata?: string;
+  };
+  maximumSizeInBytes?: number;
+  allowedContentTypes?: string[];
+}): string {}
+```
```
vvo, line 147:
```suggestion
```
This method should be called server-side, not client-side.
```

File: packages/blob/README.md [hunk 19]
```diff
@@ -128,6 +129,23 @@ async function list(options?: {
...
+
 ## Examples
```
vvo, line 153:
```suggestion
Note: This method should be called server-side, not client-side.

## Examples
```

packages/blob/README.md
correttojs: link works 👍

packages/blob/README.md
```diff
@@ -202,6 +220,124 @@ export async function POST(request: Request) {
...
+          Blob url: <a href={blob.url}>{blob.url}</a>
+        </div>
+      )}
+    </>
+  );
+}
+```
+
+```ts
+// /app/api/upload-token/route.ts
```
vvo: Should we be explicit and call this route "/app/api/generate-blob-client-token/route.ts"?

packages/blob/README.md
```diff
@@ -202,6 +220,124 @@ export async function POST(request: Request) {
...
+      },
+      maximumSizeInBytes: 10_000_000, // 10 Mb
+      allowedContentTypes: 'text/plain',
+    }),
+  });
+}
+```
+
+```ts
+// /app/api/upload-completed/route.ts
```
vvo: Should we be explicit and call this route `/app/api/file-upload-completed/route.ts`?

packages/blob/README.md
```diff
@@ -202,6 +220,124 @@ export async function POST(request: Request) {
...
+
+import { type BlobUploadCompletedEvent } from '@vercel/blob';
```
```suggestion
import { type BlobUploadCompletedEvent, verifyCallbackSignature } from '@vercel/blob';
```

double check missing imports in the README if necessary 👍

packages/blob/README.md
```diff
@@ -202,6 +220,124 @@ export async function POST(request: Request) {
...
+import { NextResponse } from 'next/server';
+
+export async function POST(request: Request): Promise<NextResponse> {
+  const body = (await request.json()) as BlobUploadCompletedEvent;
+  console.log(body);
```
vvo, line 328:
```suggestion

```

correttojs: I would keep this line (and the next one showing the output)

packages/blob/README.md
```diff
@@ -202,6 +220,124 @@ export async function POST(request: Request) {
...
+  const body = (await request.json()) as { pathname: string };
+
+  return NextResponse.json({
+    clientToken: await generateClientTokenFromReadWriteToken({
+      ...body,
+      onUploadCompleted: {
+        callbackUrl: `https://${
+          process.env.VERCEL_URL ?? ''
+        }/api/upload-completed`,
+        metadata: JSON.stringify({ foo: 'bar' }),
```
vvo: Let's avoid foo: 'bar' examples and put more real world usecases.

```suggestion
        metadata: JSON.stringify({ userId: 12345 }),
```

packages/blob/README.md
```diff
@@ -202,6 +220,124 @@ export async function POST(request: Request) {
...
+    return NextResponse.json(
+      {
+        response: 'invalid signature',
+      },
+      {
+        status: 403,
+      },
+    );
+  }
+  return NextResponse.json({
```
vvo, line 354:
```suggestion
  const metadata = JSON.stringify(body.payload.metadata);
  const blob = body.blob;
  
  console.log(metadata.userId); // 12345
  console.log(blob); // { url: '...', size: ..., uploadedAt: ..., ... }
  
  return NextResponse.json({
```

packages/blob/README.md
vvo: I've added a suggestion that does that in a more explicit way, let me know

vvo: let's keep both 👍 let's just add one line break to breathe (updating suggestions)

packages/blob/README.md
```diff
@@ -202,6 +222,124 @@ export async function POST(request: Request) {
...
+```ts
+// /app/api/upload-completed/route.ts
+
+import { type BlobUploadCompletedEvent, verifyCallbackSignature } from '@vercel/blob';
+import { NextResponse } from 'next/server';
+
+export async function POST(request: Request): Promise<NextResponse> {
+  const body = (await request.json()) as BlobUploadCompletedEvent;
+  console.log(body);
+  // { type: "blob.upload-completed", payload: { metadata: "{ foo: 'bar' }", blob: ... }}
```
vvo, line 329:
```suggestion
  // { type: "blob.upload-completed", payload: { metadata: "{ foo: 'bar' }", blob: ... }}
  
```

Adding a breather line

packages/blob/src/index.edge.test.ts
```diff
@@ -0,0 +1,87 @@
+import {
```
vvo, line 1: (for my understanding) The goal of this test file is to unit test in an edge-like Jest env?

packages/blob/src/index.node.test.ts
```diff
@@ -481,4 +489,130 @@ describe('blob client', () => {
...
+    beforeEach(() => {
+      process.env.BLOB_READ_WRITE_TOKEN = undefined;
+      mockClient = mockAgent.get(BASE_URL);
+      jest.resetAllMocks();
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-explicit-any
+      global.window = {} as any;
+    });
+    afterEach(() => {
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-explicit-any
+      global.window = undefined as any;
```
vvo: I wonder if we should run this test in jsdom Jest mode instead of relying on the node env of Jest. Thoughts?

packages/blob/src/index.edge.test.ts
correttojs: right, I'm testing the only methods that use a different implementation on edge

packages/blob/src/index.node.test.ts
```diff
@@ -481,4 +489,130 @@ describe('blob client', () => {
...
+          "contentType": "text/plain",
+          "pathname": "foo.txt",
+          "size": 12345,
+          "uploadedAt": 2023-05-04T15:12:07.818Z,
+          "url": "https://blob.vercel-storage.com/storeid/foo-id.txt",
+        }
+      `);
+    });
+
+    it('should throw when calling `put()` with an BLOB_READ_WRITE_TOKEN', async () => {
```
vvo:
```suggestion
    it('should throw when calling `put()` with a server token', async () => {
```

packages/blob/src/index.ts
```diff
@@ -1,4 +1,6 @@
 import type { Readable } from 'node:stream';
+// eslint-disable-next-line unicorn/prefer-node-protocol
+import * as crypto from 'crypto';
```
vvo, line 3: curious: was adding node:crypto failing in some ways I suppose (like not using the replacement afterwards?)

packages/blob/src/index.ts
```diff
@@ -215,13 +231,152 @@ export async function list(
...
+  }
+  const timestamp = new Date();
+  timestamp.setSeconds(timestamp.getSeconds() + 30);
+  const blobToken = getToken({ token });
+
+  const [, , , storeId = null] = blobToken.split('_');
+
+  if (!storeId) {
+    throw new Error(
+      token ? 'Invalid "token" parameter' : 'Invalid BLOB_READ_WRITE_TOKEN',
```
vvo, line 261: nice 👍 I am not sure we do so for other methods accepting token or BLOB_READ_WRITE_TOKEN

packages/blob/README.md
```diff
@@ -128,6 +129,25 @@ async function list(options?: {
 }> {}
 ```
 
+### generateClientTokenFromReadWriteToken(options)
+
+Generates a single-use token that can be used from within the client. This is useful when [uploading directly from browsers](#uploading-directly-from-browsers) to circumvent the 4MB limitation of going through a Vercel-hosted route.
+
```
vvo, line 135:
```suggestion

Once created, a client token is valid for 30 seconds. This means you have 30 seconds to initiate an upload with this token.
```

packages/blob/src/index.ts
correttojs: exactly, there seems to be an issue resolving the `browser` field in the package.json, I tried a few solutions but this seems the only one working (not that critical I guess to have the `node:` prefix)

vvo: I wonder: can we add this comment to the file so nobody tries to fix that eslint rules without knowing the impact?

packages/blob/src/index.ts
```diff
@@ -215,13 +231,152 @@ export async function list(
...
+  token,
+  signature,
+  body,
+}: {
+  token?: string;
+  signature: string;
+  body: string;
+}): Promise<boolean> {
+  const secret = getToken({ token });
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
```
vvo, line 320: If we can add a comment here too that would be nice (not obvious why we're doing this)

packages/blob/src/index.ts
```diff
@@ -215,13 +231,152 @@ export async function list(
...
+    const digest = crypto
+      .createHmac('sha256', secret)
+      .update(body)
+      .digest('hex');
+    const digestBuffer = Buffer.from(digest);
+    const signatureBuffer = Buffer.from(signature);
+
+    return (
+      digestBuffer.length === signatureBuffer.length &&
+      crypto.timingSafeEqual(digestBuffer, signatureBuffer)
```
vvo, line 332: So this is the Node.js implementation, and then the edge/browser implementation, correct?

vvo: If so, let's be a bit mode descriptive via comments maybe

packages/blob/src/index.ts
```diff
@@ -215,13 +231,152 @@ export async function list(
...
+  const verified = await globalThis.crypto.subtle.verify(
+    'HMAC',
+    await importKey(token),
+    hexToArrayByte(signature),
+    new TextEncoder().encode(body),
+  );
+  return verified;
+}
+
+function hexToArrayByte(input: string): ArrayBuffer {
```
vvo, line 344: Any insights on where we took this method from? (comment)

packages/blob/src/index.node.test.ts
correttojs: the only difference between node and the browser version is the `typeof window !== 'undefined'` check, but as we need to support the client upload it makes sense to configure jsdom

packages/blob/src/index.ts
```diff
@@ -233,6 +388,14 @@ function getToken(putOptions?: BlobCommandOptions): string {
   return process.env.BLOB_READ_WRITE_TOKEN;
 }
 
+function getApiUrl(): string {
+  return (
+    process.env.VERCEL_BLOB_API_URL ||
+    process.env.NEXT_PUBLIC_VERCEL_BLOB_API_URL ||
```
vvo, line 397: Should this env variable be added when linking stores now? Or was it already the default?

I am thinking: we may have to update our documentation/dashboard to surface this new env variable required for client uploads?

vvo: Oh wait, that's mostly for us to test things I believe, the only value needed for customers would be the default one ('https://blob.vercel-storage.com/'), 👍 

The only needed doc update would be the contributing ones we have in Notion probably/the edge-functions part

test/next/src/app/vercel/blob/api/app/upload-completed/route.ts
```diff
@@ -0,0 +1,30 @@
...
+    return NextResponse.json(
+      {
+        response: 'invalid signature',
+      },
+      {
+        status: 403,
+      },
+    );
+  }
+  return NextResponse.json({
```
vvo: Maybe add the suggestions from the README here too

packages/blob/src/index.ts
correttojs: there is a `crypto.verify` function available on node as well, but the doc only refers to the publicKey and all the examples I found for the symmetric hash use the `sign`+`timingSafeEqual` solution instead

vvo: 👍

test/next/src/app/vercel/blob/app/test/client-serverless/page.tsx
```diff
@@ -0,0 +1,44 @@
+'use client';
+import { put, type BlobResult } from '@vercel/blob';
+import { useEffect, useState } from 'react';
+
+export default function AppBodyClient({
+  searchParams,
+}: {
+  searchParams: { filename: string };
```
vvo, line 8: Is this for our automated test to pass a filename before hand? I am not following why we're reading filename from query params yet

vvo (approved):
No major review changes, this looks beautiful and will be a tremendous addition to Vercel Blob. Congrats!

test/next/src/app/vercel/blob/app/test/client-serverless/page.tsx
correttojs: it's used as test identifier (same for all the other file upload tests)

packages/blob/src/index.ts
correttojs: it's the revers operation of 
```
[...new Uint8Array(signature)]
    .map((x) => x.toString(16).padStart(2, '0'))
    .join('')
```
which is equivalent to `Buffer.from(new Uint8Array(signature)).toString('hex')`. I could not find a nice reference (probably the .NET version https://stackoverflow.com/questions/311165/how-do-you-convert-a-byte-array-to-a-hexadecimal-string-and-vice-versa) and there are different implementations. Not sure what would be a meaningful comment here, the method name seems already explicit

packages/blob/src/index.ts
correttojs: runbook updated 👍

packages/blob/src/index.ts
correttojs: it's not needed in the other methods, as `option.token` falls back to `BLOB_READ_WRITE_TOKEN`

vvo: oh right 👍

vvo: Still if I provide a token in other methods, it's cool to have the error being about the token option and not BLOB_READ_WRITE_TOKEN when it's invalid (nit, not this PR goal)

correttojs: we only parse the token in this method. In the other cases we return the message `Vercel Blob: Access denied, please provide a valid token for this resource` in case of an unauthorized response, or we throw `BLOB_READ_WRITE_TOKEN environment variable is not set. Please set it to your write token` in case there are no tokens defined.

PR #181, opened 2023-06-11
Upgrade neon and use the new direct HTTP call method

cramforce:


vvo:
@cramforce They released another version where they also speak about pooling: https://github.com/neondatabase/serverless/commits/main do you want to clean this PR and re-upgrade?

vvo:
While this PR is now passing tests, a few observations:
- do we also want to use the new HTTP method in https://github.com/vercel/storage/blob/72925fe0eb21001d1feca714d62f102e47517710/packages/postgres/src/create-client.ts#L13-L36?
- it seems the pooling client when deployed on edge will timeout after a few reloads on an API route. Details are on Slack for Vercel employees: https://vercel.slack.com/archives/C04E6JF6JN6/p1686868566133539

vvo (approved):

PR #183, opened 2023-06-11
Upgrade to latest neon driver

cramforce:
- Had to fix a kysely test due to brittle mocking
- Fixed playwright tests (were just broken)
- Fixed Vercel turbo build (was just broken)

PR #186, opened 2023-06-12
[test] Upgrade to latest neon driver

vvo:


.github/workflows/integration-tests-dev.yml
```diff
@@ -35,18 +35,9 @@ jobs:
           node-version: '18'
           cache: 'pnpm'
 
-      - uses: actions/cache@v3
```
correttojs, line 38: @vvo should we fix the playwright version instead like in https://github.com/vercel/storage/pull/179/files#diff-2e5b024378c3660c9f07ef3a5519710987316e2542186019c41e35a58fda20e1L50 ?

vvo: @correttojs See https://github.com/vercel/storage/pull/183/commits/a16bc0f019b87e522c683d56062833c3803a7b0e

vvo: and https://github.com/vercel/storage/pull/183

PR #214, opened 2023-06-15
chore(pnpm): fix lockfile and remove extraneous lockfiles

vvo:
In an effort to finally fix the Vercel build issues (https://vercel.slack.com/archives/C04E6JF6JN6/p1686778046295839?thread_ts=1686495351.821699&cid=C04E6JF6JN6)

- We had multiple lock files, we only need one I believe
- We had mentions of babel in some packages, I don't think this is required, cc @dferber90 
- We used Node 16 or Node 18 in some CI actions. Now moved to Node 18 for developers of the module.
- Upgraded pnpm to latest
- Upgraded turbo to latest
- Added a node-version, use https://github.com/Schniz/fnm
- To upgrade pnpm on your machine: `corepack prepare pnpm@8.6.2 --activate` then open a new tab


package.json
```diff
@@ -39,11 +39,11 @@
...
     "ts-jest": "29.1.0",
-    "turbo": "^1.9.3",
+    "turbo": "1.10.3",
     "typescript": "^5.0.0"
   },
-  "packageManager": "pnpm@8.4.0",
+  "packageManager": "pnpm@8.6.2",
   "engines": {
-    "node": ">=16.14"
+    "node": ">=18"
```
correttojs, line 47: we are still running node16 tests https://github.com/vercel/storage/actions/runs/5276021124/jobs/9542191279?pr=214

correttojs (approved):

package.json
vvo: Yes definitely 👍 this version is mostly for the developers iterating on this module, and for the CI actions that are not the unit test matrixes ones.

PR #216, opened 2023-06-15
allow customizing validUntil

correttojs:
the reason for exposing the `validUntil` is 
- the user can restrict or extend the `clientToken` validity depending on the use case
- the vercel api can validate the client token by calling `generateClientTokenFromReadWriteToken` with the extracted payload


packages/blob/README.md
```diff
@@ -145,6 +145,7 @@ async function generateClientTokenFromReadWriteToken(options?: {
   };
   maximumSizeInBytes?: number;
   allowedContentTypes?: string[];
+  validUntil?: number;
```
```suggestion
  validUntil?: number; // timestamp in ms
```

vvo (approved):

PR #221, opened 2023-06-16
create handleBlobUpload helpers

correttojs:
- create handleBlobUpload wrapper
- trigger the client token request from within the `put` method
- doc updated


correttojs:
@vvo should we document both the low-level interface and the wrapper? I'd prefer to just doc the wrapper
I added some notes to the [liner ticket](https://linear.app/vercel/issue/STO-666/create-next-handlers-to-simplify-the-file-upload) as we forgot about the webhooks `signature` and body string...let me know if this solution is good enough

vvo:
> I'd prefer to just doc the wrapper

Let's do that, for as long as our wrappers can handle all customization cases. Which for now is the case 👍

test/next/README.md
```diff
@@ -10,6 +10,10 @@ The root page provides links to test suites. You can visit these test suites to
 
 This also deploys to a preview branch for every PR, so you can test your code in production!
 
+## File Upload
+
+File upload is protected by the `blobUpload` cookie. Call `document.cookie="blobUpload=[process.env.BLOB_UPLOAD_SECRET]; Path=/;"` in order to set it
```
vvo: Should we use NEXT_PUBLIC_BLOB_UPLOAD_SECRET and have this env variable part of vercel.com dashboard? So when we pull the project we get it automatically.

correttojs: `BLOB_UPLOAD_SECRET` is already part of the vercel dashboard and github secrets. The idea is to don't expose it to the client, so we don't allow external people to upload files using this website

packages/blob/README.md
```diff
@@ -285,79 +275,57 @@ export default function UploadForm() {
...
-      onUploadCompleted: {
-        callbackUrl: `https://${
-          process.env.VERCEL_URL ?? ''
-        }/api/file-upload-completed`,
-        metadata: JSON.stringify({ userId: 12345 }),
+  const body = (await request.json()) as HandleBlobUploadBody;
+
+  // The request will run at least 2 times:
+  // - the first time when generating a client token, the body will contain the {type: "blob.generate-client-token"} object.
+  // - the second time once the file upload completed, the body will contain the {type: "blob.upload-completed"} object. This request will retry for 5 times in case the response returns an error status code.
```
```suggestion
  // Uploading from browsers is a three-step process:
  // - First, a specific client token is generated by this current route and sent to the browser. The request body will contain a `{ type: "blob.generate-client-token", ... }` object.
  // - Second, the file is uploaded to Vercel Blob directly from the browser, using the client token. The file content doesn't go through this route.
  // - Third, a webhook is called on this route (`onUploadCompleted`). The request body will contain a `{ type: "blob.upload-completed", ... }` object. This webhook will be retried five times in case your route doesn't reply with a 200.
```

test/next/README.md
vvo: Got it. Should the instructions say to grab the value of BLOB_UPLOAD_SECRET from their .env files?
Where should. "document.cooke=..." be pasted? I suppose in the browser console? Let's make it clear 👍

packages/blob/README.md
```diff
@@ -285,79 +275,57 @@ export default function UploadForm() {
...
-import {
-  type BlobUploadCompletedEvent,
-  verifyCallbackSignature,
-} from '@vercel/blob';
-import { NextResponse } from 'next/server';
+      onUploadCompleted: async ({ blob, metadata }) => {
+        console.log('Upload completed', blob, metadata);
+        try {
+          // Run any logic after the file upload completed
+          await db.update({ avatar: blob.url, userId: metadata.userId });
```
```suggestion
          const parsedMetadata = JSON.parse(metadata);
          await db.update({ avatar: blob.url, userId: parsedMetadata.userId });
```

?

packages/blob/README.md
```diff
@@ -285,79 +275,57 @@ export default function UploadForm() {
...
-        }/api/file-upload-completed`,
-        metadata: JSON.stringify({ userId: 12345 }),
+  const body = (await request.json()) as HandleBlobUploadBody;
+
+  // The request will run at least 2 times:
+  // - the first time when generating a client token, the body will contain the {type: "blob.generate-client-token"} object.
+  // - the second time once the file upload completed, the body will contain the {type: "blob.upload-completed"} object. This request will retry for 5 times in case the response returns an error status code.
+
+  try {
+    const jsonResponse = await handleBlobUpload({
```
vvo, line 331: I can't find where do we document the API of handleBlobUpload, the same way we document the API of https://github.com/vercel/storage/blob/fabiobenedetti/sto-666-create-next-handlers-to-simplify-the-file-upload/packages/blob/README.md#putpathname-body-options for example

test/next/README.md
correttojs: I thought `process.env.` already suggest that the value comes from the `.env` file. I can update the doc anyway

packages/blob/README.md
```diff
@@ -285,79 +275,57 @@ export default function UploadForm() {
...
+      body,
+      request,
+      onBeforeGenerateToken: async (pathname) => {
+        // On a real website, this route would be protected by authentication, see: https://nextjs.org/docs/pages/building-your-application/routing/authenticating
```
```suggestion
        // In most cases, you should authenticate users before allowing upload tokens to be generated and sent to browsers. Otherwise, you're exposing your Blob store to be an anonymous upload platform.
        // See https://nextjs.org/docs/pages/building-your-application/routing/authenticating for more informations
```

test/next/README.md
vvo: We can make this easier for contributors maybe?

For example, we could disable this behavior when running locally? I do not think we need to protect uploads when you're running a local development server. We only need to protect production via smart secrets like you did. Thoughts?

correttojs: it already works when the env is not set (I will update the doc accordingly)

vvo: Oh then maybe we do not need to say anything about this, but perhaps a section on how tests are running on GitHub.

packages/blob/src/index.ts
```diff
@@ -406,3 +238,40 @@ function mapBlobResult(blobResult: BlobMetadataApi): BlobResult {
...
+  }
+}
+
+async function checkAndRetrieveClientToken(
+  pathname: string,
+  options: PutCommandOptions,
+): Promise<void> {
+  if (!options.token && options.handleBlobUploadUrl) {
+    const { handleBlobUploadUrl } = options;
+    const url = isValidUrl(handleBlobUploadUrl)
```
vvo: isAbsoluteUrl?

packages/blob/src/index.ts
```diff
@@ -89,6 +80,8 @@ export async function put(
     throw new BlobError('access must be "public"');
   }
 
+  await checkAndRetrieveClientToken(pathname, options);
```
vvo: Can we write this code to not mutate the options object? While we do not have an eslint rule for this. This is usually not recommended:
1. It makes understanding the code harder (I had to carefully read the method to understand why we're only awaiting and not returning any result)
2. In this case, we're also mutating objects the user sends. Consider this use case:

```
// customer code
const defaultPutOptions = { handleBlobUploadUrl };

await put('file1.png', defaultPutOptions); // works
await put('file2.png', defaultPutOptions); // fails, defaultPutOptions now has a token property.
```

Recommendation:

```
const token = isClientPut(options) ? await retrieveClientToken(options) : options.token; 
```

Something similar to this might make the code more obvious and will avoid mutating objects.

packages/blob/README.md
```diff
@@ -253,19 +253,9 @@ export default function UploadForm() {
...
-            '/api/generate-blob-client-token',
-            {
-              method: 'POST',
-              body: JSON.stringify({
-                pathname: file.name,
-              }),
-            },
-          ).then((r) => r.json())) as { clientToken: string };
-
           const blobResult = await put(file.name, file, {
```
vvo, line 295:
```suggestion
          // A secured client token will be obtained automatically by making a call to `/api/upload/avatars`
          const blobResult = await put(file.name, file, {
```

packages/blob/README.md
correttojs: oh true, I forgot it!

packages/blob/README.md
correttojs: ok, I'd still prefer the `metadata: any` though 😅

packages/blob/src/index.ts
correttojs: fair point, I will make it immutable

packages/blob/README.md
```diff
@@ -129,9 +129,47 @@ async function list(options?: {
...
+      };
+}): string {}
+```
+
+Note: This method should be called server-side, not client-side.
+
 ### generateClientTokenFromReadWriteToken(options)
 
-Generates a single-use token that can be used from within the client. This is useful when [uploading directly from browsers](#uploading-directly-from-browsers) to circumvent the 4MB limitation of going through a Vercel-hosted route.
+Generates a single-use token that can be used from within the client. This method is called internally by `handleBlobUpload`.
```
correttojs, line 172: we may add an "Advanced use case" section later on, wdyt @vvo ?

vvo: Yes I think that would be useful for folks willing to go full customization as we discussed (caching token, swr, ..). Can you track this in Linear? Thanks!

packages/blob/README.md
```diff
@@ -129,9 +129,47 @@ async function list(options?: {
...
+
+Handles the requests to generate a client token and respond to the upload completed event. This is useful when [uploading directly from browsers](#uploading-directly-from-browsers) to circumvent the 4MB limitation of going through a Vercel-hosted route.
+
+```ts
+async function handleBlobUpload(options?: {
+  token?: string;
+  request?: IncomingMessage | Request;
+  onBeforeGenerateToken: (pathname: string) => Promise<{
+    allowedContentTypes?: string[];
+    maximumSizeInBytes?: number;
```
vvo, line 142: is there a default? if yes let's add it here as a comment

packages/blob/README.md
```diff
@@ -129,9 +129,47 @@ async function list(options?: {
...
+    validUntil?: number;
```
vvo: is there a default? if yes let's add it here as a comment

vvo (approved):
LGTM, should we find a few beta testers for this API, as a beta release?

packages/blob/README.md
correttojs: there is no default, we only validate this field in case it's defined

packages/blob/README.md
correttojs: https://linear.app/vercel/issue/STO-669/document-client-upload-advanced-use-case

correttojs:
> LGTM, should we find a few beta testers for this API, as a beta release?

sure, I will post an update to the #project-storage channel asking for beta users

PR #224, opened 2023-06-19
docs(contributing): add instructions on how to develop modules

vvo:
Running `pnpm dev` in the `test/next` folder is not sufficient, as packages/ are not rebuilt so you can't make changes and see them working easily.

With this change, you can now run a dev script in most packages so you can load urls from the test/next folder and have them use new code immediately. Without rebuilding.

### DIFF
--- a/packages/blob/package.json
+++ b/packages/blob/package.json
@@ -22,8 +22,9 @@
     "dist"
   ],
   "scripts": {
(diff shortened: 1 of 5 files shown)


correttojs:
nice! should we expose the `dev` script on the root too?

correttojs (approved):

vvo:
> nice! should we expose the `dev` script on the root too?

I tried to, but I could not make it work (i.e. run all "dev" scripts at once).

Issue #250, opened 2023-06-27
.set() is automatically running JSON.parse() on the result without telling you


Gbox4:
If you run:
```typescript
import { kv } from "@vercel/kv";

const a1 = "a"
await kv.set(url, a1)
const b1 = await kv.get(url)
console.log("test 1")
console.log(typeof a1)
console.log(typeof b1)

const a2 = "1"
await kv.set(url, a2)
const b2 = await kv.get(url)
console.log("test 2")
console.log(typeof a2)
console.log(typeof b2)

const a3 = "{}"
await kv.set(url, a3)
const b3 = await kv.get(url)
console.log("test 3")
console.log(typeof a3)
console.log(typeof b3)
```

you get:
```
test 1
string
string
test 2
string
number
test 3
string
object
```

Should they not all return `string`? Is the auto-parsing is the desired behavior? Where is this documented? I couldn't find it [here](https://vercel.com/docs/storage/vercel-kv/kv-reference#set).

adriancooney:
Hi @Gbox4, thanks for the report. The documentation should be clearer here. Automatic deserialization is enabled for the default client (`kv`) but if you want to disable it, you can create a separate client with `automaticDeserialization: false` via the `createClient` method. That looks like:

```js
import { createClient } from "@vercel/kv";

const customKvClient = createClient({
  url: process.env.KV_REST_API_URL,
  token: process.env.KV_REST_API_TOKEN,
  automaticDeserialization: false,
});
```

The enables the following:

```js
await kv.set("object1", { hello: "world" });

console.log(await kv.get("object1")); // { hello: 'world' }
console.log(await customKvClient.get("object1")); // '{"hello":"world"}'

await kv.set("object2", "{}");

console.log(await kv.get("object2")); // {}
console.log(await customKvClient.get("object2")); // '{}'
```

adriancooney:
Docs updated. Thanks!

PR #252, opened 2023-06-27
docs: @vercel/kv automatic deserialization

adriancooney:
Addresses https://github.com/vercel/storage/issues/250.

This PR adds docs on the `automaticDeserialization` option in `createClient`.

### DIFF
--- a/packages/kv/README.md
+++ b/packages/kv/README.md
@@ -66,6 +66,25 @@ const kv = createClient({
 await kv.set('key', 'value');
 ```
 
+### Automatic Deserialization
+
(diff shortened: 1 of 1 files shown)


correttojs (approved):

packages/kv/README.md
```diff
@@ -66,6 +66,25 @@ const kv = createClient({
 await kv.set('key', 'value');
 ```
 
+### Automatic Deserialization
+
+The default `kv` client will automatically deserializes values returned from the database via `JSON.parse`. If this behaviour is undesired, create a custom KV client via the `createClient` method with `automaticDeserialization: false`. All data will be returned as strings.
```
adriancooney:
```suggestion
The default `kv` client automatically deserializes values returned from the database via `JSON.parse`. If this behaviour is undesired, create a custom KV client via the `createClient` method with `automaticDeserialization: false`. All data will be returned as strings.
```

PR #269, opened 2023-07-05
feat(blob): Introduce API versioning

vvo:
Added VERCEL_BLOB_USE_API_VERSION so we can efficiently upgrade the vercel blob api version while sending back the right data. Instead of using only a "multibucket" flag. I also moved this param to a header so we do not mess up with the url/api params itself.

Once released, this change won't trigger anything for now. I only need it to be released so I can continue migrating the Vercel Blob API which requires vercel/blob testing. So I can trigger the new behavior from the test itself.


packages/blob/src/index.ts
```diff
@@ -229,19 +235,25 @@ export async function list(
   };
 }
 
-function getApiUrl(): string {
-  return (
+function getApiUrl(pathname = ''): string {
```
correttojs, line 238: I understand it may take more time, wouldn't it better to have `https://blob.vercel-storage.com/v2`?

correttojs (approved):

packages/blob/src/index.ts
vvo: Is there a benefit over x-api-version? Perhaps consistency with other api services?

vvo: I see that, for some reason I do not remember, we call the Vercel Blob API on put this way: blob.vercel-storage.com/pathname, so we can't have `v2` now here :D

correttojs: not that important as it's still a private api, if/when we are going to expose it, it could be a nicer interface

PR #277, opened 2023-07-12
add connection info to edge config clients

dferber90:
It's currently possible to create an Edge Config client like so

```
const client = createClient(process.env.EDGE_CONFIG)
```

But when any function receives a fully configured client as an argument, it currently has no way of e.g. knowing the id of the edge config this client connects to.

To make this possible, this PR is adding the `connection` info to the client itself.

You can now do `client.connection.id` to read the Edge Config Id.

The full info available on the connection is

```json5
{
  baseUrl: 'https://example.com/ecfg-2', // this is usually https://edge-config.vercel.com/:edgeConfigId
  id: 'ecfg-2', // the edge config id
  token: 'token-2', // the token used to connect
  type: 'external', // whether this edge config is hosted on edge-config.vercel.com or not
  version: '1', // the api version supplied in the connection string, this is always "1" at the moment
}
```

### DIFF
--- a/packages/edge-config/src/index.common.test.ts
+++ b/packages/edge-config/src/index.common.test.ts
@@ -216,6 +216,18 @@ describe('connectionStrings', () => {
       expect(typeof pkg.createClient).toBe('function');
     });
 
+    describe('connection', () => {
+      it('should contain the info parsed from the connection string', () => {
+        expect(edgeConfig.connection).toEqual({
+          baseUrl: 'https://example.com/ecfg-2',
+          id: 'ecfg-2',
+          token: 'token-2',
+          type: 'external',
+          version: '1',
+        });
+      });
+    });
+
     describe('get', () => {
       describe('when item exists', () => {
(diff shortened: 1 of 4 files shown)


AndyBitz (approved):

packages/edge-config/src/index.common.test.ts
```diff
@@ -216,6 +216,18 @@ describe('connectionStrings', () => {
       expect(typeof pkg.createClient).toBe('function');
     });
 
+    describe('connection', () => {
+      it('should contain the info parsed from the connection string', () => {
+        expect(edgeConfig.connection).toEqual({
+          baseUrl: 'https://example.com/ecfg-2',
```
dferber90, line 222: I haven't fully made up my mind whether we want to expose all these fields or whether we should start with only exposing the id <picture data-single-emoji=":thinkies:" title=":thinkies:">:thinkies:</picture>

PR #288, opened 2023-07-18
feat(blob): Move to multi-bucket by default

vvo:
BREAKING CHANGE: Some methods responses and types have been updated following a
migration Vercel did to make Vercel Blob more robust and closer to the S3 API.

Namely:
- the urls generated by Vercel Blob have moved from:
  public.blob.vercel-storage.com/zyzoioy8txfs14xe/somefile-NoOVGDVcqSPc7VYCUAGnTzLTG2qEM2.txt
to
zyzoioy8txfs14xe.blob.vercel-storage.com/somefile-NoOVGDVcqSPc7VYCUAGnTzLTG2qEM2.txt.
This change has been done transparently: all previous blob urls are
redirected to the new domain.
- .put() no more sends back `size` and `uploadedAt`, use .head() to get this
  information. Refer to the `PutBlobResult` type.
- .list() no more sends back `contentType` and `contentDisposition` on the
  `blobs` property. Refer to the `ListBlobResult` type.
- .del() response is always `null`, no other metadata is sent. del()
  doesn't inform you if the delete was succesful. If the file was here prior to
a del() call, then it's now deleted. Use head() when necessary.
- `BlobResult` type has been splitted to `PutBlobResult` for put() and
  `HeadBlobResult` for head(). Use them accordingly.

We've also reworked our README to better surface the browser-upload and
serverl-upload methods.


package.json
```diff
@@ -36,7 +36,7 @@
     "husky": "8.0.3",
     "jest": "29.6.1",
     "lint-staged": "13.2.3",
-    "prettier": "3.0.0",
+    "prettier": "2.8.8",
```
vvo, line 39: Had to downgrade prettier because of: https://github.com/prettier/eslint-plugin-prettier/issues/562

packages/blob/src/index.ts
```diff
@@ -24,38 +24,30 @@ export {
...
-
-export interface ListCommandOptions extends BlobCommandOptions {
-  limit?: number;
-  prefix?: string;
-  cursor?: string;
-}
+// This version is used to ensure that the client and server are compatible
+// The server (Vercel Blob API) uses this information to change its behavior like the
+// response format
+const BLOB_API_VERSION = 2;
```
vvo, line 30: We will first release the Vercel Blob API, which changes its behavior when api version < 2 (send back previous format).

Then we will release the package.

packages/blob/src/index.ts
```diff
@@ -115,18 +107,30 @@ export async function put(
...
+
+  return blobResult;
 }
 
-type BlobDelResult<T extends string | string[]> = T extends string
-  ? BlobResult | null
-  : (BlobResult | null)[];
+// vercelBlob.del()
 
-export async function del<T extends string | string[]>(
```
vvo, line 126: I moved away from generics to overloads because the method doesn't accepts generics.

packages/blob/README.md
```diff
@@ -56,33 +175,15 @@ async function put(
...
-  | ({
-      size: number;
-      uploadedAt: Date;
-      pathname: string;
-      contentType: string;
-      contentDisposition: string;
-      url: string;
-    } | null)[]
-> {}
+): Promise<string | string[]> {}
```
correttojs: just wondering if we want to keep the `{ url: string }` interface, which is extendable and more compatible with the previous interface

packages/blob/README.md
```diff
@@ -131,38 +230,41 @@ async function list(options?: {
...
-  token?: string;
-  request?: IncomingMessage | Request;
+  token?: string; // default to process.env.BLOB_READ_WRITE_TOKEN
+  request: IncomingMessage | Request;
   onBeforeGenerateToken: (pathname: string) => Promise<{
-    allowedContentTypes?: string[];
-    maximumSizeInBytes?: number;
-    validUntil?: number; // timestamp in ms, by default 30s
+    allowedContentTypes?: string[]; // optional, defaults to no restriction
+    maximumSizeInBytes?: number; // optional, defaults and maximum is 500MB (524,288,000 bytes)
```
correttojs, line 241: where does this limit come from? CF? 
this may change depending on the user plan, right?

vvo: Yes this comes from Cloudflare: https://developers.cloudflare.com/workers/platform/limits/#request-limits

packages/blob/README.md
vvo: 👍 for { url: string } and compatibility, but I am even wondering if we should return anything, the R2 API sends back "void" (https://developers.cloudflare.com/r2/api/workers/workers-api-reference/) and that may be the good choice. Because S3 sends back some data, but it's useless. So if the return value of delete is always useless, maybe `void` is the right thing.

packages/blob/src/index.ts
correttojs: I don't understand the exact issue but I'm fine with the overloads too

test/next/src/app/vercel/blob/handle-blob-upload.ts
```diff
@@ -36,7 +36,7 @@ export async function handleBlobUploadHandler(
         }
 
         return {
-          maxFileSize: 10_000_000,
+          maximumSizeInBytes: 10_000_000,
```
correttojs, line 39: don't we need to update `https://github.com/vercel/edge-functions` accordingly?

packages/blob/README.md
correttojs: as we support multiple deletes, it may be useful to know which files have been deleted and which threw an error (I don't think we are handling this case in the worker though)

test/next/src/app/vercel/blob/handle-blob-upload.ts
vvo: edge-functions is already accepting maximumSizeInBytes, I think at some point we renamed this and it did not rename it everywhere

packages/blob/README.md
correttojs: ok, in case we apply lower limits for hobby users we can review this comment later on

test/next/src/app/vercel/blob/handle-blob-upload.ts
correttojs: ah nice! I only checked your last PR, but indeed I see the name is `maximumSizeInBytes`👍

packages/blob/README.md
vvo: > which files have been deleted

Neither R2 nor S3 can tell us about that unfortunately. They just say "If you ask for it to be deleted, it is deleted if it exists". i.e. if you want to know if it's deleted, you only need to know if it existed before.

That's why they probably decided to return void. I am inclined to do the same as then the API response is very clear: no you can't know if it worked: it worked if the files were here

correttojs: ok, then the only reason to keep it would be for compatibility with the old api. I'd be fine either to return either `{url : string}` or just `void`

vvo: > then the only reason to keep it would be for compatibility with the old api

I'd rather optimize for new users rather than previous users of the module (because our biggest opportunity are new users). I will send void and update the Vercel Blob API too now

vvo: now sending `null` always

vvo:
Tests should work once we merge the API PR: we're failing at a CORS issue (x-api-version is not yet allowed)

packages/blob/README.md
correttojs: can't we just skip the return and set `Promise<void`> as return type?

correttojs (approved):

packages/blob/README.md
vvo: I wanted to do that but then you can't have a type DelBlobResult which is "void", so I thought "perhaps this is not a good idea and null is 'better'". Thoughts?

vvo: CleanShot 2023-07-18 at 1 05 51

hmm that's eslint maybe we can ignore this

correttojs: we can remove this type, the user wouldn't need it anymore

vvo: done and now returning void

PR #294, opened 2023-07-20
fix vite env vars

correttojs:
the client upload doesn't work on SvelteKit as the `process.env` is not available on Vite (see https://vitejs.dev/guide/env-and-mode.html)

As the 2 variables requested by the client are optional (they are used for testing against staging or v1), we can simply wrap them using a try/catch, allowing SvelteKit to load their default hardcoded value

Adding a Blob SvelteKit example to this repo for testing purposes, we will add some integration tests on a separate PR


packages/blob/src/index.ts
```diff
@@ -257,12 +257,13 @@ export async function list(
 }
 
 function getApiUrl(pathname = ''): string {
-  const baseUrl =
-    process.env.VERCEL_BLOB_API_URL ||
-    process.env.NEXT_PUBLIC_VERCEL_BLOB_API_URL ||
-    'https://blob.vercel-storage.com';
-
-  return `${baseUrl}${pathname}`;
+  let baseUrl = null;
```
dferber90, line 260: Could you add a comment here as the reasons behind this will be quite hard to understand based on the code alone

dferber90: @tcc-sejohnson could you double check this since you also prepared vite for edge config

dferber90 (commented):
random fly-by comments

packages/blob/src/index.ts
correttojs: as @vvo suggested, we could also suggest to add `define : { process.env : {} } ` to the vite config, but I think we can avoid it as the variables are used only for the local testing

dferber90: I wonder why we need the try/catch in the first place. Could you instead check for `process?.env?.XXX`?

In Edge Config we have a note in the README but I see that it might not make sense for you https://github.com/vercel/storage/blob/main/packages/edge-config/README.md#a-note-for-vite-users. I'm not that familiar with Vite but @tcc-sejohnson fixed this in Edge Config and might have some pointers for you too

packages/blob/src/index.ts
```diff
@@ -257,12 +257,15 @@ export async function list(
...
+  try {
+    // wrapping this code in a try/catch as this function is used in the browser and Vite doesn't define the process.env.
+    // As this varaible is NOT used in production, it will always default to production endpioint
```
dferber90:
```suggestion
    // As this varaible is NOT used in production, it will always default to production endpoint
```

packages/blob/src/index.ts
correttojs: the note is valid for a server access, in the client Vite injects the `import.meta.env.VITE_ABC` object.
I'm not sure how the bundlers are handling the process variables (i.e. the full `process.env.NEXT_PUBLIC_VERCEL_BLOB_API_VERSION_OVERRIDE` variable is defined, but if you try to access it dynamically `process.env[key]` it is not, also the process.env object is `{}`). So asserting on the process seems dangerous

packages/blob/src/index.ts
```diff
@@ -324,7 +327,13 @@ function shouldFetchClientToken(
 }
 
 function getApiVersionHeader(): { 'x-api-version'?: string } {
-  const versionOverride = process.env.VERCEL_BLOB_API_VERSION_OVERRIDE;
+  let versionOverride = null;
+  try {
+    // wrapping this code in a try/catch as this function is used in the browser and Vite doesn't define the process.env.
+    // As this varaible is NOT used in production, it will always default to the BLOB_API_VERSION
+    process.env.VERCEL_BLOB_API_VERSION_OVERRIDE ||
```
vvo:
```suggestion
    versionOverride = process.env.VERCEL_BLOB_API_VERSION_OVERRIDE ||
```

right? cc @correttojs

vvo (commented):
Do we need to add at least VERCEL_BLOB_READ_WRITE_TOKEN env to the vite/sveltekit project? How does it works otherwise?

vvo (changes requested):
proposed a small change to be accepted I guess

correttojs:
> Do we need to add at least VERCEL_BLOB_READ_WRITE_TOKEN env to the vite/sveltekit project? How does it works otherwise?

the `vite.config` loads the variables
```
 const env = loadEnv(mode, process.cwd(), '')
    dotenvExpand.expand({ parsed: env })
```
but they are only accessible from within the server. `VERCEL_BLOB_READ_WRITE_TOKEN` can only be accessed from 

> Do we need to add at least VERCEL_BLOB_READ_WRITE_TOKEN env to the vite/sveltekit project? How does it works otherwise?

added `.env.example`

vvo (approved):
Good for me, but this is failing

correttojs:
> Good for me, but this is failing

I moved the sveltekit example to a separate PR and fixed the eslint issue

correttojs:
@tcc-sejohnson could you please have a look at this solution? Currently it is impossible for SvleteKit users to upload blob files using the client upload flow

vvo:
@correttojs Is this PR similar to https://github.com/vercel/storage/issues/168 and https://github.com/vercel/storage/pull/169? cc @ffMathy

correttojs:
> @correttojs Is this PR similar to #168 and #169? cc @ffMathy

You should be able to consume an env variable with Vite on edge functions (there may still be an issue reading the process.env though). 
I would thread this PR as a separate case as Blob needs to work on the client too, and the 2 env we are currently accessing are optional

vvo:
> @correttojs Is this PR similar to #168 and #169? cc @ffMathy

Sorry, not the same at all

PR #295, opened 2023-07-20
docs(blob): Refer to vercel.com documentation, precise 4 MB

vvo:
- The limit is actually 4.5 MB
- Let's reference vercel.com examples instead of copy pasting examples everywhere
- Let's still keep this README API documentation as the source of truth


vvo:
To be merged once vercel.com documentation is updated

correttojs (approved):

PR #307, opened 2023-07-25
Add postgresql adapter for Kysely

### LINKED ISSUE #325: Support for kysely migration
I would be great if we can use kysely migrations. That should be achieved by disabling transactions 

Currently we have this error
```
{
  error:  [VercelPostgresError: VercelPostgresError - 'kysely_transactions_not_supported': Transactions are not supported yet.] {
  code: 'kysely_transactions_not_supported',
  name: 'VercelPostgresError'
}
```

I think this could be useful

https://kysely-org.github.io/kysely-apidoc/interfaces/DialectAdapter.html#supportsTransactionalDdl

savoiringfaire:
The original problem I had which triggered this change was this error when running migrations as documented in https://kysely.dev/docs/migrations

```
VercelPostgresKyselyError [VercelPostgresError]: VercelPostgresError - 'kysely_transactions_not_supported': Transactions are not supported yet.
```

Which we get because we haven't properly signaled to Kysely that we don't support transactions. The Kysely migrator checks `supportsTransactionalDdl` to decide whether to run migrations in a transaction or not, so this adds that.

It also overrides the default migration lock functions from https://github.com/kysely-org/kysely/blob/master/src/dialect/postgres/postgres-adapter.ts to use `pg_advisory_lock` and then unlock with `pg_advisory_unlock` since we can't rely on transactions auto-dropping now.

Fixes #325

### DIFF
--- a/packages/postgres-kysely/src/index.ts
+++ b/packages/postgres-kysely/src/index.ts
@@ -1,17 +1,60 @@
-import type { DatabaseConnection, Dialect, Driver } from 'kysely';
-import { PostgresDialect, PostgresDriver, Kysely } from 'kysely';
+import type {
+  DatabaseConnection,
+  Dialect,
+  DialectAdapter,
+  DialectAdapterBase,
+  Driver,
+  MigrationLockOptions,
+} from 'kysely';
+import {
+  PostgresDialect,
+  PostgresDriver,
+  Kysely,
+  PostgresAdapter,
+  sql,
+} from 'kysely';
 import type { Pool } from '@neondatabase/serverless';
 import { createPool } from '@vercel/postgres';
 import type { VercelPostgresPoolConfig } from '@vercel/postgres';
 import { VercelPostgresKyselyError } from './error';
 
+// Random id for our transaction lock.
+const LOCK_ID = BigInt('3853314791062309107');
+
 type VercelPostgresDialectConfig = VercelPostgresPoolConfig & {
   pool: Pool;
 };
 
+class VercelPostgresAdapter
+  extends PostgresAdapter
+  implements DialectAdapterBase
+{
+  get supportsTransactionalDdl(): boolean {
+    return false;
+  }
+
+  async acquireMigrationLock(
+    db: Kysely<any>,
+    _opt: MigrationLockOptions,
+  ): Promise<void> {
+    await sql`select pg_advisory_lock(${sql.lit(LOCK_ID)})`.execute(db);
+  }
+
+  async releaseMigrationLock(
+    db: Kysely<any>,
+    _opt: MigrationLockOptions,
+  ): Promise<void> {
+    await sql`select pg_advisory_unlock(${sql.lit(LOCK_ID)})`.execute(db);
+  }
+}
+
 class VercelPostgresDialect extends PostgresDialect implements Dialect {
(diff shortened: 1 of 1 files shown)


savoiringfaire:
Tests on this currently failing due to lack of https://github.com/kysely-org/kysely/issues/456 being officially released.

savoiringfaire:
Have un-extended the postgresql adapter and implemented the single remaining method in https://github.com/vercel/storage/pull/307/commits/20d17bed8f3e03cb387e42632a629c3e9bc614bd which has got tests passing.

volnei (commented):
Looks great

packages/postgres-kysely/src/index.ts
```diff
@@ -1,19 +1,57 @@
...
+
+  async acquireMigrationLock(db: Kysely<any>): Promise<void> {
+    await sql`select pg_advisory_lock(${sql.lit(LOCK_ID)})`.execute(db);
+  }
+
+  async releaseMigrationLock(db: Kysely<any>): Promise<void> {
+    await sql`select pg_advisory_unlock(${sql.lit(LOCK_ID)})`.execute(db);
+  }
+}
+
```
igalklebanov, line 45: not required. this should be as simple as extending `PostgresAdapter`:

```ts
class VercelPostgresAdapter extends PostgresAdapter {
  get supportsTransactionalDdl(): boolean {
    return false;
  }
}
```

packages/postgres-kysely/src/index.ts
```diff
@@ -1,19 +1,57 @@
...
+  sql,
+} from 'kysely';
 import type { Pool } from '@neondatabase/serverless';
 import { createPool } from '@vercel/postgres';
 import type { VercelPostgresPoolConfig } from '@vercel/postgres';
 import { VercelPostgresKyselyError } from './error';
 
+// Random id for our transaction lock.
+const LOCK_ID = BigInt('3853314791062309107');
+
```
igalklebanov, line 23: No need, see other comment/s..

```suggestion
```

igalklebanov (changes requested):

volnei:
Can someone explain what I can do to have this working?

vvo:
Hey there, getting back to this. We're investigating how much value our Kysely plugin provides versus a community one: https://github.com/seveibar/kysely-neon. Have any of you tried kysely-neon? We're wondering if we should deprecate our own adapter and delegate it to the community.

As you can tell, we have not been very active in this issue (apologies), which is why we're wondering if it should be owned by the community.

Let us know if https://github.com/seveibar/kysely-neon could work for you, and if not I can re-review this PR.

volnei:
> Hey there, getting back to this. We're investigating how much value our Kysely plugin provides versus a community one: https://github.com/seveibar/kysely-neon. Have any of you tried kysely-neon? We're wondering if we should deprecate our own adapter and delegate it to the community.
> 
> As you can tell, we have not been very active in this issue (apologies), which is why we're wondering if it should be owned by the community.
> 
> Let us know if https://github.com/seveibar/kysely-neon could work for you, and if not I can re-review this PR.

I agree with you since the kysely neon is an active project and cover all behaviours that vercel kysely do. Can we ensure that? Otherwise I prefer to mantain current with this fixes that I think that is the only problem.

Thanks

vvo:

PR #314, opened 2023-07-29
Remove types field in package.json files

### LINKED ISSUE #313: TypeScript complains about import when used with moduleResolution node16
When importing `@vercel/postgres` from a CommonJS module with `"moduleResolution": "Node16"` TypeScript resolves the ES module type declaration so it thinks it is importing an ES module. The result is a compilation error telling you to use a dynamic import.

A better explanation can be found [here](https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/FalseESM.md).

I found this when I tried to use `@vercel/postgres` but I'm pretty sure this happens in all packages in this repo.

I think the solution would be to remove the `"types"` field in package.json and let TypeScript infer the right file based on the name.

lukahartwig:
Fixes #313

Without the `types` field TypeScript resolves the declaration file automatically based on name and `moduleResolution` setting. In some packages I changed the `files` field to include `.d.cts` files in the published package.


vvo:
@lukahartwig this looks correct, thanks. Do you have any reference for removing the "types" at the root of the package.json? I want to be 100% sure VSCode/TypeScript will find the correct types even when moduleResolut < node16 (which will then use the `main` field). Any manual test or link to doc appreciated.

lukahartwig:
Sure, the module resolution strategy for "classic" and "node" is explained [here](https://www.typescriptlang.org/docs/handbook/module-resolution.html#module-resolution-strategies). Basically, when you import a .js file, TypeScript always looks for a .d.ts file with the same name next to it by default.

packages/blob/package.json
```diff
@@ -27,14 +27,6 @@
     "undici": "./dist/undici-browser.js",
     "crypto": "./dist/crypto-browser.js"
   },
-  "types": "./dist/index.d.ts",
-  "typesVersions": {
```
dferber90, line 31: I've removed `typesVersions` when resolving merge conflicts, but this led to an error: https://github.com/vercel/storage/actions/runs/6221448244/job/16883410004?pr=314#step:8:245

Does `typesVersions` need to stay?

dferber90: Brought this back in https://github.com/lukahartwig/storage/commit/728fd7ddb47ec0bd1c59a785a6ea0333bb3e07db

lukahartwig: I think it's trying to import `@vercel/blob/client` from a CJS context. Typescript is resolving this to `./packages/blob/client.d.ts`. A dirty hack would be to add a file there that reexports the types from the dist folder.

If you don't find a solution until then I can look at it this evening.

dferber90: It works when I bring back `typesVersions`. It's not so obvious in this PR as it fails due to you being an outside contributor, but I recreated your changes in https://github.com/vercel/storage/pull/391 and it's all green there with `typesVersions` present.

Not sure if I prefer a hack over declaring `typesVersions`, but I'm keen to hear your take on it

lukahartwig: Both are hacks, I'm fine with the `typesVersion` solution. Seems more maintainable as well 👍

packages/postgres/package.json
```diff
@@ -27,11 +26,8 @@
   },
   "main": "./dist/index.cjs",
   "module": "./dist/index.js",
```
dferber90, line 28: There seems to be another issue when trying to actually release this change, as I tested it out with a snapshot release

https://github.com/vercel/storage/actions/runs/6222845972/job/16887566314#step:8:46

```
Error: 🦋  error src/index.ts(4,28): error TS7016: Could not find a declaration file for module '@vercel/postgres'. '/home/runner/work/storage/storage/packages/postgres/dist/index.cjs' implicitly has an 'any' type.
🦋  error   Try `npm i --save-dev @types/vercel__postgres` if it exists or add a new declaration (.d.ts) file containing `declare module '@vercel/postgres';`
Error: 🦋  error src/index.ts(5,[47](https://github.com/vercel/storage/actions/runs/6222845972/job/16887566314#step:8:48)): error TS7016: Could not find a declaration file for module '@vercel/postgres'. '/home/runner/work/storage/storage/packages/postgres/dist/index.cjs' implicitly has an 'any' type.
🦋  error   Try `npm i --save-dev @types/vercel__postgres` if it exists or add a new declaration (.d.ts) file containing `declare module '@vercel/postgres';`
🦋  error 
🦋  error Error: error occured in dts build
🦋  error     at Worker.<anonymous> (/home/runner/work/storage/storage/node_modules/.pnpm/tsup@7.2.0_typescript@5.2.2/node_modules/tsup/dist/index.js:2294:26)
🦋  error     at Worker.emit (node:events:514:28)
🦋  error     at MessagePort.<anonymous> (node:internal/worker:243:53)
🦋  error     at [nodejs.internal.kHybridDispatch] (node:internal/event_target:741:20)
🦋  error     at exports.emitMessage (node:internal/per_context/messageport:23:28)
🦋  error DTS Build error
🦋  error RollupError: Failed to compile. Check the logs above.
🦋  error     at error (/home/runner/work/storage/storage/node_modules/.pnpm/rollup@3.21.2/node_modules/rollup/dist/shared/rollup.js:279:30)
🦋  error     at Object.error (/home/runner/work/storage/storage/node_modules/.pnpm/rollup@3.21.2/node_modules/rollup/dist/shared/rollup.js:2[48](https://github.com/vercel/storage/actions/runs/6222845972/job/16887566314#step:8:49)30:20)
🦋  error     at Object.error (/home/runner/work/storage/storage/node_modules/.pnpm/rollup@3.21.2/node_modules/rollup/dist/shared/rollup.js:24024:42)
🦋  error     at generateDtsFromTs (/home/runner/work/storage/storage/node_modules/.pnpm/tsup@7.2.0_typescript@5.2.2/node_modules/tsup/dist/rollup.js:7512:22)
🦋  error     at /home/runner/work/storage/storage/node_modules/.pnpm/tsup@7.2.0_typescript@5.2.2/node_modules/tsup/dist/rollup.js:7519:59
🦋  error     at _nullishCoalesce (/home/runner/work/storage/storage/node_modules/.pnpm/tsup@7.2.0_typescript@5.2.2/node_modules/tsup/dist/rollup.js:1:198)
🦋  error     at Object.transform (/home/runner/work/storage/storage/node_modules/.pnpm/tsup@7.2.0_typescript@5.2.2/node_modules/tsup/dist/rollup.js:7519:18)
🦋  error     at /home/runner/work/storage/storage/node_modules/.pnpm/rollup@3.21.2/node_modules/rollup/dist/shared/rollup.js:2[50](https://github.com/vercel/storage/actions/runs/6222845972/job/16887566314#step:8:51)29:40
```

lukahartwig: Looks like the `prepublish` script does not take into account the workspace dependency between `@vercel/postgres` and `@vercel/postgres-kysely`.

packages/blob/package.json
vvo: @dferber90 There was a file at some point but the typesVersion solution is less invading code wise. i.e. like an ie6 fix I prefer for it to be hidden as much as possible

vvo: Also the file solution is not complete, you actually need a whole folder for it to work, see https://github.com/vercel/storage/commit/b8b76332112c124c68ef4b240f9678e05c84da52 and https://antfu.me/posts/types-for-sub-modules

packages/postgres/package.json
vvo: Oh right, like we should first build vercel/postgres and then postgres-kysely, correct?

lukahartwig: Yeah, the normal release process uses `turbo run build` which takes care of that.

@dferber90 are you working on this?

dferber90: 🤦 Yeah the `pnpm build` was totally missing. Added in https://github.com/vercel/storage/pull/393 and it is working now. Thanks for pointing that out!

dferber90 (approved):

dferber90:
Appreciate your help here @lukahartwig <picture data-single-emoji=":frog-10:" title=":frog-10:">:frog-10:</picture>

PR #340, opened 2023-08-29
chore(blob): fix linting

vvo:
I have seen errors: https://github.com/vercel/storage/actions/runs/6010542607/job/16302173404?pr=318

### DIFF
--- a/packages/blob/src/index.node.test.ts
+++ b/packages/blob/src/index.node.test.ts
@@ -541,14 +541,14 @@ describe('blob client', () => {
             callbackUrl: 'https://example.com',
           },
         },
-        // eslint-disable-next-line @typescript-eslint/require-await
         onBeforeGenerateToken: async (pathname) => {
+          await Promise.resolve();
           return {
             metadata: pathname,
           };
         },
-        // eslint-disable-next-line @typescript-eslint/require-await
         onUploadCompleted: async (body) => {
+          await Promise.resolve();
           spy.call(body);
         },
       });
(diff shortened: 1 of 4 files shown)


packages/blob/src/index.node.test.ts
```diff
@@ -541,14 +541,14 @@ describe('blob client', () => {
             callbackUrl: 'https://example.com',
           },
         },
-        // eslint-disable-next-line @typescript-eslint/require-await
         onBeforeGenerateToken: async (pathname) => {
+          await Promise.resolve();
```
correttojs, line 545: instead of awaiting a fake promise, we could remove the async and return a `Promise.resolve({
            metadata: pathname,
          })`

vvo: We can, is there a difference with the fake promise? In both case we create a fake promise.

correttojs: it seems cleaner to return a promise

correttojs: oh but this is just the test 😅, fine with your solution then

vvo: Yes and also I thought "Well await Promise.resolve() could be the authentication part" so perhaps even closer to real world with the fake promise here

packages/blob/src/index.node.test.ts
```diff
@@ -559,7 +559,7 @@ describe('blob client', () => {
       });
       expect(spy).not.toHaveBeenCalled();
       expect(
-        // eslint-disable-next-line @typescript-eslint/no-unsafe-argument, @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-explicit-any
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-argument, @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-explicit-any -- Either the test is incomplete, or we're messing up with TS
```
correttojs, line 562: I think `handleBlobUpload` returns either the `clientToken` or an empty response, depending if the handler is called by the user or by the webhook
We can fix the type or simply update the comment

correttojs (approved):

PR #341, opened 2023-08-29
chore(build): try to fix pnpm lock

vvo:
We're having issues again about missing dependencies when building on Vercel, can't understand why yet. So I regenerated the pnpm file.


vvo:
no change, still failing

PR #343, opened 2023-08-29
feat(blob): add addRandomSuffix option to disable automatic ids on put

vvo:
This new options allows you to disable the addition of automatic id suffixes to
pathnames.

example:

```ts
const blobWithoutAutomaticSuffix = await put('welcome.txt', 'Hello world!', {
  access: 'public',
  addRandomSuffix: false
});

// 'https://$storeId.public.blob.vercel-storage.com/welcome.txt'
console.log(blobWithoutAutomaticSuffix.url);

const blobWithAutomaticSuffix = await put('bye.txt', 'Hello world!', {
  access: 'public'
});

// 'https://$storeId.public.blob.vercel-storage.com/bye-$someRandomId.txt'
console.log(blobWithAutomaticSuffix.url);
```

fixes STO-780

This also means you can now override blobs. We're planning to allow you to
configure the cache also, stay tuned.

### DIFF
--- a/packages/blob/src/index.ts
+++ b/packages/blob/src/index.ts
@@ -37,7 +37,8 @@ export interface BlobCommandOptions {
 export interface PutCommandOptions extends BlobCommandOptions {
   access: 'public';
   contentType?: string;
-  handleBlobUploadUrl?: string;
+  handleBlobUploadUrl?: string; // only used in the browser
+  addRandomSuffix?: boolean;
 }
 
 export interface PutBlobResult {
@@ -88,6 +89,10 @@ export async function put(
 
   if (options.contentType) {
     headers['x-content-type'] = options.contentType;
+  }
+
+  if (options.addRandomSuffix !== undefined) {
+    headers['x-add-random-suffix'] = options.addRandomSuffix ? '1' : '0';
   }
 
   const blobApiResponse = await fetch(getApiUrl(`/${pathname}`), {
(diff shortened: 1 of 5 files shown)


test/next/src/app/vercel/blob/script.ts
```diff
@@ -53,14 +54,32 @@ async function run(): Promise<void> {
 
   console.log(count, 'blobs in this store');
 
-  await Promise.all(urls.map((url) => vercelBlob.del(url as string)));
+  // await Promise.all(urls.map((url) => vercelBlob.del(url)));
```
correttojs: was it intended?

vvo: Nope, removing, thanks

packages/blob/src/index.ts
```diff
@@ -37,7 +37,8 @@ export interface BlobCommandOptions {
 export interface PutCommandOptions extends BlobCommandOptions {
   access: 'public';
   contentType?: string;
-  handleBlobUploadUrl?: string;
+  handleBlobUploadUrl?: string; // only used in the browser
+  addRandomSuffix?: boolean;
```
correttojs, line 41: wonder if we should add the `unique` word in the `addRandomSuffix` flag, to indicate that the flag allows to create a unique link per upload

vvo: `addRandomSuffix` was the top proposal API name from a ChatGPT conversation :D

I would say it's a fine name

correttojs (approved):

PR #344, opened 2023-08-30
feat(blob): Allow configuring cache-control max age

vvo:
This PR adds a new option on `put` and client token generation named `cacheControlMaxAge?: number`.

Blobs are cached for one year by default (edge, browsers). This new option allows you to configure this behavior.


packages/blob/src/index.ts
```diff
@@ -39,6 +39,7 @@ export interface PutCommandOptions extends BlobCommandOptions {
   contentType?: string;
   handleBlobUploadUrl?: string; // only used in the browser
   addRandomSuffix?: boolean;
+  cacheControlMaxAge?: number;
```
correttojs, line 42:
```
{ 
access: 'public';
contentType?: string;
addRandomSuffix?: boolean;
cacheControlMaxAge?: number;
} | 
{
access: 'public'; 
handleBlobUploadUrl?: string; 
}
```

it would be nice to have 2 interfaces for client and server, we didn't consider the server-only options while implementing the client upload 😓

correttojs: now we cannot force the client upload interface in case of client requests, unless we create a new `clientPut` method

vvo: Decision:
- We're reviewing this PR as it is
- I'll open a second PR to clean up the types and add a clientPut method.

Reasoning: The smart .put() that acts differently when we detect the parameter handleBlobUploadUrl should be a different method, as its responsibility is entirely different than regular .put(). clientPut() is the companion of handleBlobUpload (server).

But it doesn't have to be a different import path. Ultimately all of these methods can be called server or client side. Later on we will allow to use list() with a client token which would limit the path (prefixes) you can list (think: list the photos you uploaded without hitting the database).

correttojs (approved):

PR #347, opened 2023-09-01
feat(blob): Move client methods to @vercel/blob/client + renaming

vvo:
This pull request introduces BREAKING CHANGES. Mostly, we've been trying to separate client and server needs better. While ensuring we only export what we currently think is useful to you.

We've moved and renamed client-related utilities, including the ones to generate client tokens, to a separate entry file: `import { put, upload, handleUpload, generateClientTokenFromReadWriteToken} from '@vercel/blob/client'`.

Here are the most important BREAKING CHANGES:
- `BlobCommandOptions` type is no longer exported. This is an internal utility type. Other internal types were removed also but nobody was using them
- handleBlobUpload has moved to client/`handleUpload`
- For client (browser) uploads, please use client/`upload` instead of `put`. Also, the `handleBlobUploadUrl` option has been renamed to `handleUploadUrl`.
- `verifyCallbackSignature` is no more exported if you think you would like to use it, open an issue

---

This PR allows `clientPayload` during `upload,` so you can attach
metadata to the file uploaded to Vercel Blob. This metadata will be sent as
`tokenPayload` by default in uploading completed callbacks. And you can also enrich
it/configure it/verify it.

Usage:
```ts
// client.ts
import { upload } from '@vercel/blob/client';

upload(
  'file.txt',
  {
    access: 'public',
    handleUploadUrl: '/api/upload',
    clientPayload: JSON.stringify({ postId: 123 })
  }
);
```

```ts
const jsonResponse = await handleUpload({
  body,
  request,
  onBeforeGenerateToken: async (pathname, clientPayload) => {

    return {
       tokenPayload: clientPayload // this is the default, ideally you would verify the clientPayload (format, authorization) before generating the token, so no files are uploaded without first verifying the client claims
    };
  },
  onUploadCompleted: async ({ blob, tokenPayload }) => {
    console.log('blob upload completed', blob, tokenPayload);
  }
});
```

An example use-case of this feature would be to allow editing the images of a
blog post. To do so you would use `clientPayload: JSON.stringify({ postId: 345
})` in the browser. Then, either in onBeforeGenerateToken or in onUploadCompleted
you would verify that the client making the call is allowed to modify this
postId (authorization) and update your database accordingly.

BREAKING CHANGES:
- `metadata` in `onBeforeGenerateToken` and `onUploadCompleted` has been renamed to `tokenPayload`


packages/blob/src/index.ts
```diff
@@ -122,6 +116,96 @@ export async function put(
...
+    | string
+    | Readable
+    | Blob
+    | ArrayBuffer
+    | FormData
+    | ReadableStream
+    | File,
+  options: ClientPutCommandOptions
+): Promise<PutBlobResult> {
+  if (!pathname) {
```
correttojs: should we throw if `typeof window === undefined`? not 100% sure that the request comes from the client, but it could still be useful

packages/blob/src/index.ts
```diff
@@ -76,12 +75,7 @@ export async function put(
     throw new BlobError('access must be "public"');
   }
 
-  const token = shouldFetchClientToken(options)
-    ? await retrieveClientToken({
-        handleBlobUploadUrl: options.handleBlobUploadUrl,
-        pathname,
-      })
-    : getToken(options);
+  const token = getTokenFromOptionsOrEnv(options);
```
correttojs: same here, could we throw in case `typeof window` is defined?

packages/blob/src/index.ts
```diff
@@ -122,6 +116,96 @@ export async function put(
...
 
+// vercelBlob.clientPut()
+// This is a wrapper that will fetch the client token for you and then upload the file
+export interface ClientPutCommandOptions extends BlobCommandOptions {
+  access: 'public';
+  contentType?: string;
+  handleClientUploadUrl: string;
+}
+
+export async function clientPut(
```
correttojs: can we try to create a `client/index.ts` and re-export only this file?
I know we said it's fine to expose the `clientPut` function, but think that it would be harder to move it once we'll release the public beta

packages/blob/README.md
```diff
@@ -31,11 +31,30 @@ async function put(
     // `token` defaults to process.env.BLOB_READ_WRITE_TOKEN on Vercel
     // and can be configured when you connect more stores to a project
     // or using Vercel Blob outside of Vercel
-    // During browser uploads, a token is automatically generated by calling the route from the `handleBlobUploadUrl` option
+    // You could also pass a client token here if you generated it yourself with `generateClientTokenFromReadWriteToken`
```
correttojs: it seems a bit confusing this point, why don't we allow the client upload only from the `clientPut`? if the user wants to pass the clientToken explicitly, we can create a type like `{handleClientUploadUrl} | {clientToken}`

vvo: I see `clientPut` as a convenience method companion to the handleBlobUpload (server) method. But nothing prevents someone from using the put() method directly from the frontend. Either via a regular token or a client token. There's a good probability some people might just want to roll their custom client upload doing so, and I don't want to artificially limit them doing so.

packages/blob/src/index.ts
vvo: I would not do so (same reasons as my comment here: https://github.com/vercel/storage/pull/347/files#r1313037483). put() can be used client or server side, the fact that we provide a wrapper (clientPut) that does the token exchange automatically should not prevent anyone to build their own client upload wrappers if they feel limited/constrained by ours (handleBlobUpload).

packages/blob/src/index.ts
vvo: When (= why) would we need to move some methods to /client?

As of today, I am unsure if there will be a time when we will need to move clientPut() to a different import path: all methods of the blob package can be called on the frontend. For example, if you wanted to allow list/delete/head to be called from the frontend, enforcing a specific prefix for which paths can be listed (think: per-user/per-x blob folder).

packages/blob/README.md
correttojs: so it's possible to send a `cacheControlMaxAge` option from the client by manually creating a token. The option would be then ignored by the worker 🤔

packages/blob/src/index.ts
vvo: Same as https://github.com/vercel/storage/pull/347/files#r1313037483, we should not artificially limit where put() can be used only because we provide a convenience clientPut() method. Some people will not want to use it.

packages/blob/README.md
vvo: > so it's possible to send a cacheControlMaxAge option from the client by manually creating a token

Yes, by using a read write token on the client (not recommended but 🤷 ).
Yes, by creating a client token and using `put()`. In this case, this is where we could (maybe) add runtime checks and (for sure) server-side checks on what is received.

correttojs: then I don't see the advantages of adding a `clientPut`, the additional typescript interface `{handleClientUploadUrl} | {cacheControlMaxAge,...}` would be enough...are there other reasons to create a `clientPut`?

vvo: Yes, I think it clarifies the intent and behavior of the put method when you want to use the client upload helpers versus having a method that changes its behavior based on its arguments.

packages/blob/src/client-upload.node.test.ts
```diff
@@ -0,0 +1,186 @@
+import type { IncomingMessage } from 'node:http';
```
vvo: I moved all index.test.ts client upload tests to this file, that's it

test/next/tsconfig.json
```diff
@@ -10,7 +10,7 @@
     "noEmit": true,
     "esModuleInterop": true,
     "module": "esnext",
-    "moduleResolution": "node",
+    "moduleResolution": "Node",
```
vvo, line 13: this doesn't change anything, just TS automatically set it with a capital Node somehow if you change it

packages/blob/src/index.ts
vvo: I moved methods to client when they are related to client uploads in general

packages/blob/src/client.ts
```diff
@@ -145,30 +179,31 @@ export function getPayloadFromClientToken(
...
   type: (typeof EventTypes)['generateClientToken'];
   payload: { pathname: string; callbackUrl: string };
 }
-export interface BlobUploadCompletedEvent {
+interface UploadCompletedEvent {
   type: (typeof EventTypes)['uploadCompleted'];
   payload: {
+    // TODO @Fabio: is this correct? I guess it should be a PutBlobResult and not HeadBlobResult unless
+    // we're doing an extra head() call somewhere?
     blob: HeadBlobResult;
```
correttojs: I think at the time we implemented this feature the 2 interfaces were exactly the same
Now the event receives 
```
 const blob: BlobPutApiResult | BlobPutApiResultV0 =
    apiVersion < 2
      ? // we're faking previous response format for the vercel/blob module not to throw
        {
          url: encodeURI(`${baseUrlForBlobs}/${key}`),
          pathname,
          contentType,
          contentDisposition,
          uploadedAt: new Date().toISOString(),
          size: parseInt(request.headers.get('content-length') ?? '0', 10),
        }
      : {
          url: encodeURI(`${baseUrlForBlobs}/${key}`),
          pathname,
          contentType,
          contentDisposition,
        };
```
so `PutBlobResult` should be the correct one

packages/blob/src/put.ts
```diff
@@ -0,0 +1,125 @@
...
+} from './helpers';
+import { type PutCommandOptions } from '.';
+
+const putOptionHeaderMap = {
+  cacheControlMaxAge: 'x-cache-control-max-age',
+  addRandomSuffix: 'x-add-random-suffix',
+  contentType: 'x-content-type',
+};
+
+export interface PutBlobResult {
```
correttojs, line 28: just an idea for a separate PR...it would be nice to consume these interfaces from the worker, to ensure client and server uses the same fields

packages/blob/src/put.ts
```diff
@@ -0,0 +1,125 @@
...
+  pathname: string;
+  contentType: string;
+  contentDisposition: string;
+}
+
+export type PutBlobApiResponse = PutBlobResult;
+
+type PartialBy<T, K extends keyof T> = Omit<T, K> & Partial<Pick<T, K>>;
+
+export function createPutMethod<
```
correttojs, line 39: nice interface!

correttojs (approved):
tested the client and server upload, all good 👍
great job!

packages/blob/src/client.ts
vvo: Updated 👍

packages/blob/src/put.ts
vvo: Indeed, I have been reluctant to do so because then it ties both services very tightly (for example: you cannot easily add a new property to edge-functions unless you update the npm package).

packages/blob/src/put.ts
vvo: 👍

PR #355, opened 2023-09-06
chore(blob): new release

vvo:


vvo:
This is non-breaking and for @Schniz

Schniz (approved):
🥰

PR #362, opened 2023-09-11
feat(blob): add `clientPayload` and `tokenPayload`

vvo:
This PR allows `clientPayload` during `upload,` so you can attach
metadata to the file uploaded to Vercel Blob. This metadata will be sent as
`tokenPayload` by default in uploading completed callbacks. And you can also enrich
it/configure it/verify it.

Usage:
```ts
// client.ts
import { upload } from '@vercel/blob/client';

upload(
  'file.txt',
  {
    access: 'public',
    handleUploadUrl: '/api/upload',
    clientPayload: JSON.stringify({ postId: 123 })
  }
);
```

```ts
const jsonResponse = await handleUpload({
  body,
  request,
  onBeforeGenerateToken: async (pathname, clientPayload) => {

    return {
       tokenPayload: clientPayload // this is the default, ideally you would verify the clientPayload (format, authorization) before generating the token, so no files are uploaded without first verifying the client claims
    };
  },
  onUploadCompleted: async ({ blob, tokenPayload }) => {
    console.log('blob upload completed', blob, tokenPayload);
  }
});
```

An example use-case of this feature would be to allow editing the images of a
blog post. To do so you would use `clientPayload: JSON.stringify({ postId: 345
})` in the browser. Then, either in onBeforeGenerateToken or in onUploadCompleted
you would verify that the client making the call is allowed to modify this
postId (authorization) and update your database accordingly.

BREAKING CHANGES:
- `metadata` in `onBeforeGenerateToken` and `onUploadCompleted` has been renamed to `tokenPayload`

TODO before release:
- [ ] modify edge-functions to accept tokenPayload AND metadata
- [ ] test on staging for real
- [ ] update front PR documentation

### DIFF
--- a/packages/blob/src/client.ts
+++ b/packages/blob/src/client.ts
@@ -369,7 +369,7 @@ export interface GenerateClientTokenOptions extends BlobCommandOptions {
   pathname: string;
   onUploadCompleted?: {
     callbackUrl: string;
-    tokenPayload?: string | null;
+    tokenPayload?: string;
   };
   maximumSizeInBytes?: number;
   allowedContentTypes?: string[];
--- a/packages/blob/src/client.ts
+++ b/packages/blob/src/client.ts
@@ -8,7 +8,7 @@ import { fetch } from 'undici';
 import type { BlobCommandOptions } from './helpers';
 import { BlobError, getTokenFromOptionsOrEnv } from './helpers';
 import { createPutMethod } from './put';
-import type { PutBlobResult } from '.';
+import type { PutBlobResult } from './put';
 
 // client.put()
 export interface ClientPutCommandOptions {
@@ -49,6 +49,7 @@ export interface UploadOptions {
   access: 'public';
   contentType?: string;
   handleUploadUrl: string;
+  clientPayload?: string;
 }
 
 export const upload = createPutMethod<UploadOptions>({
@@ -80,6 +81,7 @@ export const upload = createPutMethod<UploadOptions>({
     const clientToken = await retrieveClientToken({
       handleUploadUrl: options.handleUploadUrl,
       pathname,
+      clientPayload: options.clientPayload,
     });
     return clientToken;
   },
@@ -95,7 +97,7 @@ async function importKey(token?: string): Promise<CryptoKey> {
   );
 }
 
-async function signPayload(
+export async function signPayload(
   payload: string,
   token: string
 ): Promise<string | undefined> {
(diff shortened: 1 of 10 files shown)


packages/blob/src/client.ts
```diff
@@ -232,8 +236,12 @@ export async function handleUpload({
   const type = body.type;
   switch (type) {
     case 'blob.generate-client-token': {
-      const { pathname, callbackUrl } = body.payload;
-      const payload = await onBeforeGenerateToken(pathname);
+      const { pathname, callbackUrl, clientPayload } = body.payload;
+      const payload = await onBeforeGenerateToken(pathname, clientPayload);
+      // TODO @Fabio: the next line `?? null` previously, which always made sure that tokenPayload was `null`
```
vvo: cc @correttojs the previous code was : `metadata: payload.metadata ?? null` which ensured the JSON was always `null`, while other types suggest metadata is always either undefined or a string, any reason why?

correttojs: I remember there was something with the metadata serialization. not sure if the `undefined` would cause any issue while serializing/deserializing the client token.
We can add a generateToken unit test passing an empty payload eventually

correttojs:
I just noticed we need to update `edge-functions` to send the `tokenPayload` instead of the `metadata` to the webhook (not the webhook itself luckily)

packages/blob/src/client.ts
```diff
@@ -232,8 +236,12 @@ export async function handleUpload({
...
+      // Why so?
+      const tokenPayload = payload.tokenPayload ?? clientPayload;
```
correttojs, line 243: that's a bit confusing indeed...we can generate a payload from the server (through `onBeforeGenerateToken`, think the case we need to send the session Id) or from the client (directly through the `clientPayload` for any unsecure data). Which of the two values should have precedence? should we replace this logic by a function, which potentially alows to combine server and client payload?

correttojs: oh I see now, we are passing the `clientPayload` to the `onBeforeGenerateToken`. That should be good then (not sure what's the reason of your comment then).

May not be clear for the user that  `onBeforeGenerateToken` can return a tokenPayload...maybe an optional async `onGeneratePayload(clientToken)` would be more clear?

correttojs (approved):

packages/blob/src/client.ts
vvo: Added a unit test, fine

packages/blob/src/client.ts
vvo: Discussed in call: we keep it this way. The naming internally is not obvious between the various uses of the word "payload" but that's on us.

PR #363, opened 2023-09-12
add stale-if-error semantics

dferber90:
Adds stale-if-error semantics to the Edge Config SDK.

This will reuse the previous response whenever the response to an Edge Config read has status code 500, 502, 503 or 504. This list of codes to apply stale-if-error semantics is from [Cache Control on MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control).It will also reuse the previous response whenever fetch throws.

If no previous response exists the behavior is unchanged and the SDK will still throw.

There is currently no max age for stale responses. 

To decide:
- whether we want to introduce a maxAge / ttl for stale-if-error responses -> Yes
- whether we want to make stale-if-error behavior configurable via options in `createClient(process.env.EDGE_CONFIG, options)` -> Yes

### DIFF
--- a/packages/edge-config/src/index.ts
+++ b/packages/edge-config/src/index.ts
@@ -17,7 +17,7 @@ import type {
   EdgeConfigValue,
   EmbeddedEdgeConfig,
 } from './types';
-import { createFetchWithCachedResponse } from './utils/fetch-with-cached-response';
+import { fetchWithCachedResponse } from './utils/fetch-with-cached-response';
 
 export {
   parseConnectionString,
@@ -135,12 +135,11 @@ interface EdgeConfigClientOptions {
    *
    * Any response with a status code of 500, 502, 503, or 504 is considered an error.
    *
-   * Pass `Infinity` to always use stale cached entries on error.
-   * Pass a negative number to turn disable stale-if-error semantics.
+   * Pass a negative number, 0, or false to turn disable stale-if-error semantics.
    *
-   * The time is supplied in seconds. Defaults to `Infinity`.
+   * The time is supplied in seconds. Defaults to one week (`604800`).
    */
-  staleIfError: number;
+  staleIfError: number | false;
 }
 
 /**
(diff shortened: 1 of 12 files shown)


packages/edge-config/src/utils/fetch-with-cached-response.ts
```diff
@@ -9,46 +11,115 @@ interface ResponseWithCachedResponse extends Response {
...
+ * it implmenets stale-if-error semantics.
+ */
 export async function fetchWithCachedResponse(
   url: string,
   options: FetchOptions = {}
 ): Promise<ResponseWithCachedResponse> {
-  const { headers: customHeaders = new Headers(), ...customOptions } = options;
-  const authHeader = customHeaders.get('Authorization');
-  const cacheKey = `${url},${authHeader || ''}`;
+  try {
```
AndyBitz: I think this try-catch block should be removed.

It logs the error and then throws it, I think we only want to throw it and pass it on, to not pollute the output even if the error will be handled.

dferber90: 🤦 Good catch! I added this for debugging purposes and forgot to remove it

AndyBitz (approved):

packages/edge-config/src/index.ts
```diff
@@ -127,6 +127,22 @@ function getConnection(connectionString: string): Connection | null {
...
+   * is generated locally - for example due to a connection error.
+   *
+   * Any response with a status code of 500, 502, 503, or 504 is considered an error.
+   *
+   * Pass `Infinity` to always use stale cached entries on error.
+   * Pass a negative number to turn disable stale-if-error semantics.
+   *
+   * The time is supplied in seconds. Defaults to `Infinity`.
+   */
+  staleIfError: number;
```
dferber90: I made `staleIfError` configurable and pass it down to `fetchWithCachedResponse`, but I wonder if we should instead pass it using a request header directive. This would work perfectly for this case, but might not work for other kinds of options. I'm still going back and forth, but leaning towards passing it using a request header instead.

PR #372, opened 2023-09-13
disable calculated versions

dferber90:
When turning this option off all snapshot releases will start with `0.0.0-` instead of using a calculated version.

I'm not sure what we prefer, but opening the PR now so we can decide.

```diff
-0.4.0-579bd13f016c7de43a2830340634b3948db358b6-20230913164912
+0.0.0-579bd13f016c7de43a2830340634b3948db358b6-20230913164912
```

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -9,6 +9,6 @@
   "updateInternalDependencies": "patch",
   "ignore": [],
   "snapshot": {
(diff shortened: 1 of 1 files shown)


dferber90:
Let's keep it the way it is for now, so that we also get an idea of what the version would ultimately end up being if the PR got merged.

PR #387, opened 2023-09-18
feat(blob): Add `cacheControl` property to head() calls

vvo:


packages/blob/src/index.ts
```diff
@@ -62,10 +62,11 @@ export interface HeadBlobResult {
   pathname: string;
   contentType: string;
   contentDisposition: string;
+  cacheControl: string;
 }
 
 interface HeadBlobApiResponse extends Omit<HeadBlobResult, 'uploadedAt'> {
-  uploadedAt: string;
+  uploadedAt: string; // when receiving data from our API, uploadedAt is a string
```
correttojs, line 69: shouldn't it be partial then?

correttojs (approved):

packages/blob/src/index.ts
vvo: We do receive updatedAt but it's a string, and when we return it to the customer it becomes a Date so it's omit to change its type

PR #399, opened 2023-09-20
introduce `toAbsoluteUrl(..)`

### LINKED ISSUE #398: Relative `handleUploadUrl` paths in @vercel/blob being incorrectly joined
**Link:** https://github.com/daryl-cecile/bug-report-next-relative-upload-url

**Repro:**
1. Start application in dev `bun dev`
2. Navigate to home page and open dev tools
3. Click 'Account'
4. Click on the button to upload a file
5. Notice failure in console `TypeError: URL is not valid or contains user credentials.`

**Expects vs Actual**
I expect the upload to happen without failures. Instead I get a `TypeError: URL is not valid or contains user credentials.` error in the console. On closer inspection it seems that the url being created in the `retrieveClientToken(..)` function is not correctly created:

image

It looks like the url is simply being appended to `window.location.origin` if it is not absolute.

I would expect to be able to pass in paths relative to where I am without having to create the full path myself (ie. `./api/upload` instead of `location.origin + location.pathname + '/api/uploads'`) as the component doing the upload may live in different pages/routes.

darylcecile:
Introduces `toAbsoluteUrl(..)` to allow for `handleUploadUrl` to support relative paths.

Fixes https://github.com/vercel/storage/issues/398

### DIFF
--- a/packages/blob/src/client.ts
+++ b/packages/blob/src/client.ts
@@ -292,7 +292,7 @@ async function retrieveClientToken(options: {
   const { handleUploadUrl, pathname } = options;
   const url = isAbsoluteUrl(handleUploadUrl)
     ? handleUploadUrl
-    : `${window.location.origin}${handleUploadUrl}`;
+    : toAbsoluteUrl(handleUploadUrl);
 
   const event: GenerateClientTokenEvent = {
     type: EventTypes.generateClientToken,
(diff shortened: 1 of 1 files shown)


vvo (approved):

vvo:
Thank you @daryl-cecile

PR #406, opened 2023-09-26
[blob] review error codes

correttojs:
review error codes

### DIFF
--- a/packages/blob/README.md
+++ b/packages/blob/README.md
@@ -57,7 +57,7 @@ async function del(
 
 ### head(url, options)
 
-Get the metadata of a blob by its full URL. Returns `null` when the blob does not exist.
+Get the metadata of a blob by its full URL. Throws a `BlobNotFoundError` when the blob does not exist.
 
 ```ts
 async function head(
@@ -73,7 +73,7 @@ async function head(
   contentDisposition: string;
   url: string;
   cacheControl: string;
-} | null> {}
+}> {}
 ```
 
 ### list(options)
@@ -206,6 +206,8 @@ All methods of this module will throw if the request fails for either:
 
 - missing parameters
 - bad token or token doesn't have access to the resource
+- suspended Blob store
+- Blob file or Blob store not found
 - or in the event of unknown errors
 
 You should acknowledge that in your code by wrapping our methods in a try/catch block:
@@ -279,10 +281,9 @@ export default defineConfig(({ mode }) => {
 import { put } from '@vercel/blob';
 + import { BLOB_TOKEN } from '$env/static/private';
 
-const kv = await head("filepath", {
+const blob = await head("filepath", {
 -  token: '<token>',
 +  token: BLOB_TOKEN,
 });
 
-await kv.set('key', 'value');
 ```
--- a/packages/blob/src/helpers.ts
+++ b/packages/blob/src/helpers.ts
@@ -1,5 +1,8 @@
 // common util interface for blob raw commands, not meant to be used directly
 // this is why it's not exported from index/client
+
+import { type Response } from 'undici';
+
 export interface BlobCommandOptions {
   token?: string;
 }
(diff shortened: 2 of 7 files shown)


packages/blob/src/helpers.ts
```diff
@@ -32,12 +35,64 @@ export class BlobAccessError extends Error {
...
+type BlobApiError = { error?: { code: BlobApiErrorCodes } } | undefined;
+
+export async function validateBlobApiResponse(
+  response: Response
+): Promise<Error | undefined> {
+  if (response.status !== 200) {
+    if (response.status < 500) {
+      let data: BlobApiError;
+      try {
+        data = (await response.json()) as BlobApiError;
```
vvo: I wonder if this is 100% correct: when you use .json() on a response, then you cannot use it again later on.
Example: if we receive 404 for a non-existing blob, then we use .json in this method, then we may try again in the head method, and this will throw (and TypeScript won't warn us).

Even if our code may acknowledge for that, I think `validateBlobApiResponse` should either throw OR return the json data to be used in the rest of the program, which would be easier to follow and less error-prone.

Thoughts?

correttojs: good point, I will update this code 👍

vvo: something like:

```
const apiResponse = await validateBlobApiResponse(...)
```

packages/blob/src/helpers.ts
```diff
@@ -32,12 +35,65 @@ export class BlobAccessError extends Error {
...
+export async function validateBlobApiResponse(
+  response: Response
+): Promise<unknown> {
+  if (response.status !== 200) {
+    if (response.status < 500) {
+      let data: unknown;
+      try {
+        data = await response.json();
+      } catch {
+        // ignore
```
vvo: Curious: why ignore instead of immediately throwing an unknown error here? if we can't response.json then we know it's an unknown error

correttojs: what if the 404 json response body is empty?

vvo: From what I can see, the 404 JSON response is not empty in our API, it always contains JSON data. If it's empty then that would be an unknown error.

packages/blob/src/helpers.ts
```diff
@@ -32,12 +35,65 @@ export class BlobAccessError extends Error {
   }
 }
 
+export class BlobNotFoundError extends Error {
```
vvo: Can we rename this to BlobStoreNotFoundError instead? and make sure `not_found` is `store_not_found` here and in the API.

packages/blob/src/helpers.ts
```diff
@@ -32,12 +35,65 @@ export class BlobAccessError extends Error {
...
+  constructor() {
+    super('Vercel Blob: This store does not exist');
+  }
+}
+
+export class BlobSuspendedError extends Error {
```
vvo: Should we rename to BlobStoreSuspendedError?

packages/blob/src/helpers.ts
correttojs: ok, I updated the test so it matches the worker response (we may need to review this response though it's irrelevant for now as we don't read it)

packages/blob/src/helpers.ts
```diff
@@ -32,12 +35,65 @@ export class BlobAccessError extends Error {
...
 export class BlobUnknownError extends Error {
   constructor() {
     super('Vercel Blob: Unknown error, please visit https://vercel.com/help');
   }
 }
 
+type BlobApiErrorCodes =
+  | 'suspended_store'
+  | 'forbidden'
+  | 'not_found'
```
vvo, line 63: store_not_found?

packages/blob/src/helpers.ts
correttojs: yes, btw I know we decided to return `null` in case the `head` cannot find the blob, but I'd consider throwing an error instead, wdyt?

packages/blob/src/helpers.ts
correttojs: that's the one defined from our vercel api and the worker

vvo: From the Vercel API, when you try to authenticate a token, not_found without store_ prefix makes sense, because you're inquiring about an auth token related to a store. From the Vercel BLOB API, you can have either:
- store is not found
- or store is found but blob is not found

I believe they should have two different codes:
- not_found (blob, the resource you're targeting)
- store_not_found

i.e. on the DX side we need to provide details (am I trying to query a non-existing store, or is the blob just not here?)

packages/blob/src/helpers.ts
vvo: I see S3 throws when an object doesn't exist.
I see R2 doesn't throw when an object doesn't exist.

Since our methods can throw in general (for multiple reasons), I would be fine if we were to throw on object not found 👍 What do you think?

correttojs: I agree, throwing a specific error seems more meaningful for the user

correttojs:
@vvo I tested it with the updated worker api, everything works as expected now

vvo:
Nice @correttojs, can we now:
- Update the README accordingly
- pnpm changeset (minor) in the PR with a clear indication like `BREAKING CHANGE: head() will now throw instead of returning null when the blob does not exists`
- Update the front documentation

packages/blob/README.md
```diff
@@ -279,10 +281,9 @@ export default defineConfig(({ mode }) => {
 import { put } from '@vercel/blob';
 + import { BLOB_TOKEN } from '$env/static/private';
 
-const kv = await head("filepath", {
```
vvo, line 282: Oh wow how come I missed that for so long lol

vvo (approved):

PR #430, opened 2023-10-19
docs(blob): add jsdoc comments to all exposed functions

luismeyer:


packages/blob/src/client.ts
```diff
@@ -46,12 +46,34 @@ export const put = createPutMethod<ClientPutCommandOptions>({
 // upload()
 // This is a client-side wrapper that will fetch the client token for you and then upload the file
 export interface UploadOptions {
+  /**
+   * Whether the blob should be publicly accessible. Support for private blobs is planned.
+   */
   access: 'public';
+  /**
+   * Defines the media type of the blob. By default, this value is inferred from the pathname.
```
```suggestion
   * Defines the media type of the blob. By default, this value is inferred from the pathname. Sent as the 'content-type' header when downloading a blob.
```

Not sure we should be that specific, and if so it should also map what we have on the vercel.com documentation I guess.

packages/blob/src/client.ts
```diff
@@ -46,12 +46,34 @@ export const put = createPutMethod<ClientPutCommandOptions>({
...
 
+/**
+ * Uploads a blob into your store from the client.
+ * Detailed documentation can be found here: https://vercel.com/docs/storage/vercel-blob/using-blob-sdk#client-uploads
+ *
+ * If you want to upload from your server instead, check out the documentation for the put operation: https://vercel.com/docs/storage/vercel-blob/using-blob-sdk#upload-a-blob
+ *
+ * @param pathname - The pathname to upload the blob to. This includes the filename.
+ * @param body - The contents of your blob. This has to be a supported fetch body type https://developer.mozilla.org/en-US/docs/Web/API/fetch#body.
+ * @param options - Additional options.
```
vvo, line 75: Should we/can we document said options here?

vvo: Or is the editor autocomplete experience gonna fix this?

packages/blob/src/helpers.ts
```diff
@@ -4,9 +4,36 @@
...
+   * Whether the blob should be publicly accessible. Support for private blobs is planned.
+   */
+  access: 'public';
+  /**
+   * Adds a random suffix to the filename.
+   * @defaultvalue true
+   */
+  addRandomSuffix?: boolean;
+  /**
+   * Defines the media type of the blob. By default, this value is inferred from the pathname.
```
vvo:
```suggestion
   * Defines the content type of the blob. By default, this value is inferred from the pathname. Sent as the 'content-type' header when downloading a blob.
```

packages/blob/src/copy.ts
```diff
@@ -8,17 +8,6 @@ import {
   validateBlobApiResponse,
 } from './helpers';
 
-export interface CopyCommandOptions extends BlobCommandOptions {
```
vvo, line 11: Perhaps we should do this in a different PR?

And I think it's still valuable to have different interfaces for every command. So we can branch when needed. We should think about these types as public APIs (they are).

There should be a PutCommandOptions and a CopyCommandOptions that are separate on the public API level. Then internally they can be built via `extends` if necessary so we keep things DRY when doable.

packages/blob/src/client.ts
luismeyer: i stumbled upon this as well. Describing it the JSDoc way with destructuring or type annotations (https://jsdoc.app/tags-param.html) is prohibited by eslint. I think the autocomplete kinda fixes it because I added the comments inside the options type, but if you can think of something more descriptive here, I'd be happy to change it 👍

vvo: Can you screenshot the experience (i.e. how does it looks in vscode)? Thanks!

luismeyer: ## This is visible on hovering `.put`:
Screenshot 2023-10-23 at 16 52 14

--- 

## This is shown in the autocomplete dropdown:
Screenshot 2023-10-23 at 16 52 27

--- 

## And the same is shown without forcing the autocomplete dropdown by just start typing:
Screenshot 2023-10-23 at 16 52 35

vvo (approved):
LGTM! Nice work

PR #435, opened 2023-10-23
feat: make ListBlobResultBlob type publicly accessible

### LINKED ISSUE #413: Provide Typescript type for single blob result from ListBlobResult
Can we make `ListBlobResultBlob[]` a public type so that we can use it when iterating over a list of blob results? `ListBlobResult` is exported which contains `blobs: ListBlobResultBlob[]` but sadly  `ListBlobResultBlob` is not importable.

luismeyer:
closes https://github.com/vercel/storage/issues/413

### DIFF
--- a/.changeset/fuzzy-mayflies-jump.md
+++ b/.changeset/fuzzy-mayflies-jump.md
@@ -1,0 +1,5 @@
+---
+"@vercel/blob": patch
+---
+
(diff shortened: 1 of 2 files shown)


correttojs (approved):

correttojs:
just to check, do we need any doc update?

luismeyer:
> just to check, do we need any doc update?

you mean in the front docs? I don't think so, at least I don't see a place where it would be needed.

PR #454, opened 2023-10-31
chore(deps): upgrade test deps

vvo:


test/next/tsconfig.json
```diff
@@ -10,7 +10,7 @@
     "noEmit": true,
     "esModuleInterop": true,
     "module": "esnext",
-    "moduleResolution": "Node",
+    "moduleResolution": "Bundler",
```
correttojs, line 13: I didn't know about this option 👀

.github/workflows/unit-tests.yml
```diff
@@ -19,7 +19,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [16.x, 18.x, 20.x]
+        node-version: [18.x, 20.x]
```
correttojs, line 22: we still have 16 available in vercel, do we want to drop these tests? are they failing?

correttojs (approved):

.github/workflows/unit-tests.yml
vvo: Yep Next.js dropped Node 16 support in 14.x

test/next/tsconfig.json
vvo: This is the default of Next.js tsconfig.json nowadays. BTW "Node" means module resolution is Node 10x lol..

PR #457, opened 2023-11-01
read edge config using websocket while in dev

dferber90:
This is a tiny optimization relevant for development only.

At the moment reading edge config in preview and production is orders of magnitude faster than reading edge config in development, since we can't apply as many optimizations when reading edge config in development.

While in development we were sending a network request on every read. This only takes ~100ms to resolve, but we can actually do better. Instead of sending a HTTP request for every read, we now initiate a websocket subscription every time an edge config client is created in development.

This optimization brings subsequent reads during development down to under 1ms, while still ensuring updates are propagated. We chose this approach over something like stale-while-revalidate as it means values will always be up to date.

The downside of this approach is that it introduces a slight divide between the logic running in development and production.

We might also want to add an option or env var to disable this optimization.

### DIFF
--- a/packages/edge-config/src/index.ts
+++ b/packages/edge-config/src/index.ts
@@ -137,7 +137,7 @@ export function createClient(
       {
         // @ts-expect-error -- this is defined in undici which is used during dev
         headers: { authorization: connection.token },
-      }
+      },
     );
 
     ws.addEventListener('message', (event) => {
--- a/packages/edge-config/src/index.ts
+++ b/packages/edge-config/src/index.ts
@@ -61,6 +61,16 @@ async function consumeResponseBodyInNodeJsRuntimeToPreventMemoryLeak(
   await res.arrayBuffer();
 }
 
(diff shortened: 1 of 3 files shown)


packages/edge-config/src/index.ts
```diff
@@ -114,10 +124,48 @@ export function createClient(
...
+  if (
+    process.env.NODE_ENV === 'development' &&
+    process.env.EDGE_CONFIG_DISABLE_WEBSOCKET !== '1' &&
+    typeof WebSocket !== 'undefined'
+  ) {
+    const ws = new WebSocket(
+      `wss://edge-config.vercel.com/websocket/connect?edgeConfigId=${connection.id}`,
+      {
+        // @ts-expect-error -- this is defined in undici which is used during dev
+        headers: { authorization: connection.token },
```
dferber90, line 139: It's not yet clear whether this will be the final way to authenticate, so this PR is still in draft mode until this is settled.

PR #465, opened 2023-11-02
read edge config using swr semantics while in dev

dferber90:
This is a tiny optimization relevant for development only.

At the moment reading edge config in preview and production is orders of magnitude faster than reading edge config in development, since we can't apply the same optimizations when reading edge config in development.

While in development we were sending a network request on every read. This only takes ~20-100ms to resolve, but we can actually do better. Instead of sending and awaiting a HTTP request for every read, we now use swr semantics to return a value early while refreshing it in the background.

This optimization brings subsequent reads during development down to under 5ms, while still ensuring updates are propagated after the next refresh. 

The downside of this approach is that it introduces a slight divide between the logic running in development and production.

Disabling this behavior is possible by setting the `EDGE_CONFIG_DISABLE_DEVELOPMENT_SWR` env var. This swr behavior is also disabled by default if `NODE_ENV` is anything other than `development` to prevent this from running on preview or production deployments.

### DIFF
--- a/packages/edge-config/src/index.ts
+++ b/packages/edge-config/src/index.ts
@@ -121,7 +121,7 @@ export function createClient(
    */
   const shouldUseSwr =
     process.env.NODE_ENV === 'development' &&
-    process.env.EDGE_CONFIG_DISABLE_WEBSOCKET !== '1';
+    process.env.EDGE_CONFIG_DISABLE_DEVELOPMENT_SWR !== '1';
 
   const api: Omit<EdgeConfigClient, 'connection'> = {
     async get<T = EdgeConfigValue>(key: string): Promise<T | undefined> {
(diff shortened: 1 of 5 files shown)


AndyBitz (approved):

packages/edge-config/src/index.ts
```diff
@@ -114,10 +115,18 @@ export function createClient(
...
 
-  return {
-    connection,
+  /**
+   * While in development we use SWR-like behavior for the api client to
+   * reduce latency.
+   */
+  const shouldUseSwr =
+    process.env.NODE_ENV === 'development' &&
+    process.env.EDGE_CONFIG_DISABLE_WEBSOCKET !== '1';
```
dferber90: Rename

PR #471, opened 2023-11-05
fix(postgres-kysely): Disable Postgres transactions

aleksrutins:
Right now, when you try to use Kysely transactions, you get an error. However (https://github.com/kysely-org/kysely/issues/644), if you implement `supportsTransactionalDdl` to return `false` on your Adapter class, Kysely will not attempt to use transactions. This PR does that.


aleksrutins:
Oh, sorry, I didn't see #307. Apologies, closing.

PR #474, opened 2023-11-06
adds per-request caching of reads and automatic batching

dferber90:
Todo
- [ ] Object.freeze does not freeze deeply
- [ ] process.nextTick usage leads to warning and breaks integration test since the api route seems to become unresponsive as it waits forever (https://vercel-storage-next-integration-test-suite-7ixx7cbn9.vercel.app/vercel/edge-config/app/edge)

This PR adds per-request caching of Edge Config items to the SDK. It also batches requests to Edge Config.

1. Caches reads per request when hosted on Vercel
2. Automatically batches reads if distinct keys are read in the same event loop tick
3. If the current request has already seen a `getAll()` call then any `get()` or `has()` calls will just resolve from `getAll()`'s result items for the current request
4. if `getAll()` was called before and `get()` or `has()` is called for keys which do not exist, then we also resolve immediately
5. If a read to `getAll()` is in flight (has not resolved yet, but was started) then any subsequent calls to `get()` or `has()` will await that result instead of starting a new request

## Optimized cases

**All of the following cases will now result in a single read to Edge Config as they get batched and deduped. Each code block, when executed, would result in a single edge config read going over the network.**

This also means you can now kick of a `getAll()` at the beginning of your request handler without awaiting if you know that you will read edge config later. This will prime the SDK with the right values and most likely let you resolve from cache later on. Using this strategy is only recommended when your Edge Config has few keys and is small in size. Otherwise you might end up unnecessarily transferring data which you will not read for every request.

Note that this optimization only works when deployed to Vercel. It will not kick in when developing locally or hosting elsewhere.

#### same key read multiple times

```js
await get("a")
await get("a")
```

#### `getAll()` of all items mixed with `get()` and `has()`

```js
await getAll()
await get("a-key-part-of-the-items") // a key which exists
await get("a-key-not-part-of-the-items")  // a key which does not exist
await has("a-key-part-of-the-items") // a key which exists
await has("a-key-not-part-of-the-items")  // a key which does not exist
```

#### `get()` followed by `getAll()` on individual key

```js
await get("a")
await getAll(["a"])
```

#### `get()` followed by `has()`

```js
await get("a")
await has("a")
```

#### `get()`s kicked off in parallel get batched automatically

```js
await Promise.all([get("a"), get("b")])
```

## Other cases

When loading a key and then loading multiple keys, any loaded keys will not be fetched again

```js
await get("a")
await getAll(["a", "b"])
```

The `getAll` call in the snippet above will only fetch `"b"` since `"a"` was already fetched.

## Unoptimized cases

The following cases can not be batched or deduped yet

#### subsequent `get()`s of distinct keys

```js
await get("a")
await get("b")
```

Use `await getAll(["a", "b"])` instead.

#### `has()` followed by `get()`

```js
await has("a")
await get("a")
```

#### `get()` followed by `getAll()`

```js
await get("a")
await getAll()
```

## Development

This PR also overhauls the stale-while-revalidate semantics used during local development. Details in https://github.com/vercel/storage/pull/496.


packages/edge-config/src/utils/create-loaders.ts
```diff
@@ -0,0 +1,338 @@
...
+              if (isDynamicServerError(error)) throw error;
+              throw new Error(ERRORS.NETWORK);
+            },
+          );
+        }),
+      );
+    },
+    {
+      batchScheduleFn,
+      // disable batching until we have an endpoint to btch evaluate "has" calls,
```
AndyBitz:
```suggestion
      // disable batching until we have an endpoint to batch evaluate "has" calls,
```

packages/edge-config/src/utils/create-loaders.ts
```diff
@@ -0,0 +1,338 @@
...
+        {
+          headers: new Headers(headers),
+          cache: 'no-store',
+        },
+      ).then(
+        async (res) => {
+          if (res.ok) {
+            return (await res.json()) as EdgeConfigItems;
+          }
+          await consumeResponseBodyInNodeJsRuntimeToPreventMemoryLeak(res);
```
AndyBitz: Do we need to await these? Or could we just `void` them?

packages/edge-config/src/index.ts
```diff
@@ -85,6 +45,60 @@ interface EdgeConfigClientOptions {
...
+
+    const loaders = createLoaders(options);
+    loadersInstanceCache.set(requestContext, loaders);
+    return loaders;
+  }
+
+  // there is no requestConext so we can not cache loader instances per request,
+  // so we return a new instance every time effectively disabling dataloader
+  // batching and caching
+  return createLoaders(options);
```
AndyBitz: Do you know how much overhead this causes if new instances are created each time?

AndyBitz (approved):

packages/edge-config/src/utils/create-loaders.ts
dferber90: good shout, fixed in f26318ea71eae611b0a2edf6b36c0aa0163816ac

packages/edge-config/src/index.ts
dferber90: Not sure, but data loader is intended to be used this way, so it should be acceptable

dferber90: I took a few measurements and it's negligible

```
// first call
getLoadersInstance: 0.268ms
// subsequent calls
getLoadersInstance: 0.038ms
getLoadersInstance: 0.065ms
getLoadersInstance: 0.089ms
getLoadersInstance: 0.05ms
getLoadersInstance: 0.027ms
getLoadersInstance: 0.024ms
```

packages/edge-config/src/utils/create-loaders.ts
```diff
@@ -0,0 +1,486 @@
...
+const cachedJsonParseTraced = trace(
+  (edgeConfigId: string, content: string) => {
+    const cached = jsonParseCache.get(edgeConfigId);
+    if (cached) return cached;
+
+    const parsed = jsonParseTraced(content) as unknown;
+
+    // freeze the object to avoid mutations of the return value of a "get" call
+    // from affecting the return value of future "get" calls
+    jsonParseCache.set(edgeConfigId, Object.freeze(parsed));
```
AndyBitz: That wouldn't prevent nested objects from being changed, right? So we'd need to recursively freeze it.

dferber90:
Closing this as our v1 release is coming up and the best bits of this PR have been moved over independently.

PR #476, opened 2023-11-06
feat(blob): add  list `mode` param to merge blob folder

luismeyer:
This PR introduces a new `mode` param to the `list` options object. Based on this param the response either contains only blobs or in addition to that also list of folders.

### DIFF
--- a/.changeset/metal-pigs-hear.md
+++ b/.changeset/metal-pigs-hear.md
@@ -1,0 +1,5 @@
+---
+"@vercel/blob": minor
+---
+
+Adds a new `mode` parameter to the list command options. Mode controls the way our API handles folders and blob files.
--- a/.changeset/metal-pigs-hear.md
+++ b/.changeset/metal-pigs-hear.md
@@ -2,4 +2,4 @@
 "@vercel/blob": minor
 ---
 
-Adds a new `mode` parameter to the list command options. Mode controls the way our API handles folders and blob files.
+Adds a new `mode: folded | expanded (default)` parameter to the list command options. When you pass `folded` to `mode`, then we automatically fold all files belonging to the same folder into a single folder entry. This allows you to build file browsers using the Vercel Blob API.
--- a/packages/blob/src/index.ts
+++ b/packages/blob/src/index.ts
@@ -44,6 +44,7 @@ export type {
   ListBlobResultBlob,
   ListBlobResult,
   ListCommandOptions,
+  ListFoldedBlobResult as FoldedListBlobResult,
 } from './list';
 export { list } from './list';
 
--- a/packages/blob/src/index.ts
+++ b/packages/blob/src/index.ts
@@ -44,7 +44,7 @@ export type {
   ListBlobResultBlob,
   ListBlobResult,
   ListCommandOptions,
-  ListFoldedBlobResult as FoldedListBlobResult,
+  ListFoldedBlobResult,
 } from './list';
 export { list } from './list';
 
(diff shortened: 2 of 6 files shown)


packages/blob/src/index.ts
```diff
@@ -112,93 +112,24 @@ export async function head(
   return mapBlobResult(headResult);
 }
 
-// vercelBlob.list()
```
luismeyer: i moved all the list stuff into a new file. I can undo this, but it felt like this file was growing too large

vvo: Would you be open to do so in a first PR, then add your PR on top of it?
If you feel like moving all methods I am fine too, it's a good idea.

This will provide a better blame once merged into main.

luismeyer: good idea 👍 PR: https://github.com/vercel/storage/pull/478

.changeset/heavy-grapes-travel.md
```diff
@@ -0,0 +1,5 @@
+---
+"@vercel/blob": patch
```
vvo: This is a minor version since this is a new feature (we use semver if you want to learn more)

.changeset/metal-pigs-hear.md
```diff
@@ -0,0 +1,5 @@
+---
+"@vercel/blob": minor
+---
+
+Adds a new `mode` parameter to the list command options. Mode controls the way our API handles folders and blob files.
```
vvo:
```suggestion
Adds a new `mode: folded | expanded (default)` parameter to the list command options. When you pass `folded` to `mode`, then we automatically fold all files belonging to the same folder into a single folder entry. This allows you to build file browsers using the Vercel Blob API.
```

packages/blob/src/list.ts
```diff
@@ -20,40 +20,59 @@ export interface ListBlobResult {
...
  * Detailed documentation can be found here: https://vercel.com/docs/storage/vercel-blob/using-blob-sdk#list-blobs
  *
  * @param options - Additional options for the request.
  */
-export async function list(
-  options?: ListCommandOptions,
-): Promise<ListBlobResult> {
+export async function list<
+  M extends 'expanded' | 'folded' | undefined = undefined,
+>(options?: ListCommandOptions<M>): Promise<ListCommandResult<M>> {
```
vvo, line 75: Can you screenshot the VSCode experience from your preview package, when using mode: folded or mode: expanded or no option? I am specifically interested in knowing when does VSCode proposes `.folders`

packages/blob/src/list.ts
```diff
@@ -64,6 +83,10 @@ export async function list(
   if (options?.cursor) {
     listApiUrl.searchParams.set('cursor', options.cursor);
   }
+  if (options?.mode) {
+    listApiUrl.searchParams.set('mode', options.mode);
```
vvo, line 87: Can we add one unit test to check the new option is passed down to the API? we already have list tests

packages/blob/src/list.ts
luismeyer: ## No options, same as before
Screenshot 2023-11-07 at 12 03 54

## No options or `mode: expanded` doesn't expose folders
Screenshot 2023-11-07 at 12 08 33

## All autocomplete fields
Screenshot 2023-11-07 at 12 06 58

##  Autocomplete for mode union
Screenshot 2023-11-07 at 12 07 50

##  `mode: folded` exposes folders
Screenshot 2023-11-07 at 12 08 44

##  And we also still have jsdoc for all the props
Screenshot 2023-11-07 at 12 12 05
Screenshot 2023-11-07 at 12 12 44

vvo: really nice work, thanks

vvo (approved):

vvo:
@luismeyer let's add documentation on vercel.com right away too (SDK doc)

PR #494, opened 2023-11-17
feat: Better templating for `@vercel/postgres`

elliott-with-the-longest-name-on-github:
# TODO:

- [ ] Security audit
- [ ] API approval
- [ ] README

# What problem does this fix?

Right now, SQL templating is very limited. This is cool:

```ts
// this works great!
const { rows } = await sql`SELECT * FROM users WHERE user_id = ${userId}`;
```

... but it falls apart quickly:

```ts
// nope!
const { rows } = await sql`SELECT * FROM users WHERE user_id IN (${userIds})`;
```

It completely falls apart for any sort of dynamic query building, such as is often required for bulk `UPDATE` / `INSERT` statements. This adds the following API surface area:

| Utility       | Signature                                                                                                                            | Purpose                                                                                                                                                                                                                                                       |
| ------------- | ------------------------------------------------------------------------------------------------------------------------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `debug`       | `(strings: TemplateStringsArray, ...values: unknown[]) => [string, unknown[]]`                                                       | Returns a tuple of your query and its parameters -- useful if you need to log the final compiled query to see what's going wrong.                                                                                                                  |
| `fragment`    | `(strings: TemplateStringsArray, ...values: unknown[]) => [string, unknown[]]`                                                       | Has the same API as `query`, but returns a `TqlFragment` node which can be recursively nested within itself and included in a top-level `query`.                                                                                                              |
| `identifiers` | <code>(ids: string &#124; string[]) => TqlIdentifiers</code>                                                                         | Accepts a list of strings, escapes them, and inserts them into the query as identifiers (table or column names). Identifiers are safe and easy to escape, unlike query values! Will also accept a single identifier, for convenience.                         |
| `list`        | `(parameters: unknown[]) => TqlList`                                                                                                 | Accepts a list of anything and inserts it into the query as a parameterized list. For example, `[1, 2, 3]` would become `($1, $2, $3)` with the original values stored in the parameters array.                                                               |
| `values`      | `(entries: ValuesObject) => TqlValues`, where `ValuesObject` is `{ [columnName: string]: unknown }` or an array of that object type. | Accepts an array of records (or, for convenience, a single record) and builds a VALUES clause out of it. See the example below for a full explanation.                                                                                                        |
| `set`         | `(entry: SetObject) => TqlSet`, where `SetObject` is `{ [columnName: string]: unknown }`.                                            | Accepts a record representing the SET clause, and returns a parameterized SET clause. See example below for a full explanation.                                                                                                                               |
| `unsafe`      | `(str: string) => TqlTemplateString`                                                                                                 | Accepts a string and returns a representation of the string that will be inserted VERBATIM, UNESCAPED into the compiled query. Please, for all that is good, it's in the name -- this is unsafe. Do not use it unless you absolutely know your input is safe. |

The API for `sql` remains unchanged.

## An example, using the above (obviously somewhat contrived to show how various API works):

```ts
import { sql, fragment, identifiers, values } from '@vercel/postgres';

const newUser = { first_name: 'vercelliott', family_name: 'smith' };

// these could come from a column picker
const columns = identifiers(['first_name', 'family_name']);

const insertStatement = fragment`
  INSERT INTO users ${values(newUser)}
  RETURNING family_name
`;

const { rows: familyMembers } = await sql`
WITH new_user AS (
  ${insertStatement}
)

SELECT ${columns} 
FROM users
WHERE family_name IN (SELECT family_name FROM new_user);
`
```

[Full docs for the TQL spec](https://www.npmjs.com/package/@sejohnson/tql?activeTab=readme)

### DIFF
--- a/packages/postgres/src/index.ts
+++ b/packages/postgres/src/index.ts
@@ -1,11 +1,22 @@
 import type { QueryResult, QueryResultRow } from '@neondatabase/serverless';
 import { type VercelPool, createPool } from './create-pool';
-import type { Primitive } from './sql-template';
+import { tql } from './sql-template';
 
 export * from './create-client';
 export * from './create-pool';
 export * from './types';
 export { postgresConnectionString } from './postgres-connection-string';
+
+export const {
+  fragment,
+  identifier,
+  identifiers,
+  list,
+  values,
+  set,
+  unsafe,
+  query: debug,
+} = tql;
 
 let pool: VercelPool | undefined;
 
@@ -46,7 +57,7 @@ export const sql = new Proxy(
 ) as unknown as VercelPool &
   (<O extends QueryResultRow>(
     strings: TemplateStringsArray,
-    ...values: Primitive[]
+    ...values: unknown[]
   ) => Promise<QueryResult<O>>);
 
 export const db = sql;
--- a/packages/postgres/src/index.ts
+++ b/packages/postgres/src/index.ts
@@ -9,7 +9,6 @@ export { postgresConnectionString } from './postgres-connection-string';
 
 export const {
   fragment,
-  identifier,
   identifiers,
   list,
   values,
--- a/packages/postgres/src/sql-template.ts
+++ b/packages/postgres/src/sql-template.ts
@@ -1,3 +1,31 @@
-import { init, PostgresDialect } from '@sejohnson/tql';
+import { type CompiledQuery, init, PostgresDialect } from '@sejohnson/tql';
+import { VercelPostgresError } from './error';
 
-export const tql = init({ dialect: PostgresDialect });
+const { query, ...rest } = init({ dialect: PostgresDialect });
+
+export const tql = {
+  query: (
+    strings: TemplateStringsArray,
+    ...values: unknown[]
+  ): CompiledQuery => {
+    try {
+      return query(strings, ...values);
+    } catch (e) {
+      // this is for backwards-compatibility; there's no reason we can't remove it in a future major
+      if (
+        typeof e === 'object' &&
+        e !== null &&
+        'code' in e &&
+        typeof e.code === 'string' &&
+        e.code === 'untemplated_sql_call'
+      ) {
+        throw new VercelPostgresError(
+          'incorrect_tagged_template_call',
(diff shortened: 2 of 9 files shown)


packages/postgres/src/sql-template.ts
```diff
@@ -1,31 +1,31 @@
+import { type CompiledQuery, init, PostgresDialect } from '@sejohnson/tql';
 import { VercelPostgresError } from './error';
 
-export type Primitive = string | number | boolean | undefined | null;
```
elliott-with-the-longest-name-on-github, line 3: the type of `values` is now `unknown[]` instead of `Primitive[]`, but this is backwards-compatible because all `Primitive` values are assignable to `unknown`.

packages/postgres/src/index.ts
```diff
@@ -1,12 +1,22 @@
...
 export { postgresConnectionString } from './postgres-connection-string';
 
+export const {
+  fragment,
+  identifiers,
+  list,
+  values,
+  set,
+  unsafe,
+  query: debug,
```
luismeyer, line 17: maybe this is just me, but i expected `debug` to behave differently. Here is how:
1. I run into a bad query and get some neon db error i don't understand
2. I go into the code and switch `sql` with `debug`
3. get the query logged or thrown

instead the return type of `debug` is different to `sql`, so TS is unhappy and also my runtime error changed because, for example i can't read `.rows` on the result of `debug`. Does it make sense to wrap this with a little function to enhance DX event more?

elliott-with-the-longest-name-on-github: Hmm, that might be a good call... log the query and the parameters (with a check on something like `process.env.DEVELOPMENT` or something) and still send the query to the database...

packages/postgres/src/sql-template.ts
```diff
@@ -1,31 +1,31 @@
...
+const { query, ...rest } = init({ dialect: PostgresDialect });
```
AlejandroYanes, line 4: From the docs of `@sejohnson/tql` I think the helper functions coming out of `init({ dialect: PostgresDialect })` are all upper-case but they are being exported in `packages/postgres/src/index.ts` as lower-case

vvo:

PR #504, opened 2023-11-27
feat: Add TQL package (published as `@vercel/tql`)

elliott-with-the-longest-name-on-github:
See details in the package README and in #494. 

POST-MERGE:
- [ ] Replace `@sejohnson/tql` with `@vercel/tql` in #494 
- [ ] Declare a Changesets link for `@vercel/tql` => `@vercel/postgres` after (or when) #494 merges
- [ ] Write a Tinybird adapter?

### DIFF
--- a/packages/tql/README.md
+++ b/packages/tql/README.md
@@ -114,6 +114,30 @@ const [q, params] = query`SELECT * FROM users ${whereClause}`;
 
 Fragments can be nested recursively, so the possibilities are endless.
 
+If you need to combine a group of fragments, you can use `fragment.join`, which works a bit like Python's `String.join` API:
+
+```ts
+const maxAge = 30;
+const minAge = 25;
+const firstName = undefined;
+
+const filters = [];
+if (maxAge) filters.push(fragment`age < ${maxAge}`);
+if (minAge) filters.push(fragment`age > ${minAge}`);
+if (firstName) filters.push(fragment`firstName = ${firstName}`);
+
+let whereClause = fragment``;
+if (filters.length > 0) {
+  joinedFilters = fragment` AND `.join(...filters);
+  whereClause = fragment`WHERE ${joinedFilters}`;
+}
+const [q, params] = query`SELECT * FROM users ${whereClause};`;
+// output: [
+//   'SELECT * FROM users WHERE age < $1 AND age > $2;',
+//   [30, 25]
+// ]
+```
+
 ### Identifiers
 
 Need to dynamically insert identifiers?
--- a/packages/tql/README.md
+++ b/packages/tql/README.md
@@ -4,7 +4,7 @@
 pnpm i @vercel/tql
 ```
 
-TQL (Template-SQL -- unfortunately, T-SQL is already taken!) is a lightweight library for writing SQL in TypeScript:
+TQL is a lightweight library for writing SQL in TypeScript:
 
 ```ts
 import { init, PostgresDialect } from '@vercel/tql';
--- a/packages/tql/README.md
+++ b/packages/tql/README.md
@@ -1,0 +1,166 @@
+# What is it?
+
+```sh
+pnpm i @vercel/tql
+```
+
+TQL (Template-SQL -- unfortunately, T-SQL is already taken!) is a lightweight library for writing SQL in TypeScript:
+
+```ts
+import { init, PostgresDialect } from '@vercel/tql';
+
+const { query, fragment, identifiers, list, values, set, unsafe } = init({
+  dialect: PostgresDialect,
+});
+
+const [q, params] = query`SELECT * FROM users`;
+// output: ['SELECT * FROM users', []]
+```
+
+Its API is simple -- everything starts and ends with `query`, which returns a tuple of the compiled query string and parameters to pass to your database.
+
+## A Primer on Tagged Templates
+
+What is a tagged template in JavaScript? Quite simply, it's a function with the following signature:
+
+```ts
+function taggedTemplate(
+  strings: TemplateStringsArray,
+  ...values: unknown[]
+): any {
+  // just like any regular function, you can return anything from here!
+}
+
+const parameterName = 'strings';
+const result = taggedTemplate`The stringy parts of this are split on the "holes" and passed in as the "${variableName}" array!`;
+// strings = ['The stringy parts of this are split on the "holes" and passed in as the ", " array!"]
+// values: ['strings']
+```
+
+This is good news for SQL! The values in `strings` can _only_ come directly from your code (as in, you as a developer definitely wrote them), which means they're always safe. We can then view the "holes" (anything between `${}`) as a doorway through which more-exotic, potentially-unsafe things can be safely passed into your query.
+
+But how do we know what to do with the special values passed through the "doorways"? Let's break it down. We need two pieces of semantic information:
+
+- The developer's intent (what should the query compiler do with this information?)
+- The actual data being passed through
+
+This is trivially easy to do with classes -- if we only allow subclasses of a parent class (called `TqlNode`) to be passed in, then we can do something simple like this when building our query tree:
+
+```ts
+if (!value instanceof TqlNode) {
+  return new TqlParameter(value);
+}
+return value;
+```
+
+Now we know all of the values we have are instances of our parent class, we can use which specific subclass they are to determine what to do with them. For example, if the node is a `TqlParameter` node, we would add a `$1` (or whichever number it should have) into the query string and pass the `node.data` value into the parameters array.
+
+## API
+
+To start, you'll need to initialize the query compiler:
+
+```ts
+import { init, PostgresDialect } from '@vercel/tql';
+
+const tql = init({ dialect: PostgresDialect });
+```
+
+Missing your dialect? Feel free to open a PR -- they're pretty easy to write!
+
+Below, you can see the utilities returned from `init`, but here's a summary table:
+
+| Utility       | Signature                                                                                                                            | Purpose                                                                                                                                                                                                                                                       |
+| ------------- | ------------------------------------------------------------------------------------------------------------------------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `query`       | `(strings: TemplateStringsArray, ...values: unknown[]) => [string, unknown[]]`                                                       | The top-level query object. Returns a tuple of the SQL query string and an array of parameters. Pass both of these to your database driver.                                                                                                                   |
+| `fragment`    | `(strings: TemplateStringsArray, ...values: unknown[]) => [string, unknown[]]`                                                       | Has the same API as `query`, but returns a `TqlFragment` node which can be recursively nested within itself and included in a top-level `query`.                                                                                                              |
+| `identifiers` | <code>(ids: string &#124; string[]) => TqlIdentifiers</code>                                                                         | Accepts a list of strings, escapes them, and inserts them into the query as identifiers (table or column names). Identifiers are safe and easy to escape, unlike query values! Will also accept a single identifier, for convenience.                         |
+| `list`        | `(parameters: unknown[]) => TqlList`                                                                                                 | Accepts a list of anything and inserts it into the query as a parameterized list. For example, `[1, 2, 3]` would become `($1, $2, $3)` with the original values stored in the parameters array.                                                               |
+| `values`      | `(entries: ValuesObject) => TqlValues`, where `ValuesObject` is `{ [columnName: string]: unknown }` or an array of that object type. | Accepts an array of records (or, for convenience, a single record) and builds a VALUES clause out of it. See the example below for a full explanation.                                                                                                        |
+| `set`         | `(entry: SetObject) => TqlSet`, where `SetObject` is `{ [columnName: string]: unknown }`.                                            | Accepts a record representing the SET clause, and returns a parameterized SET clause. See example below for a full explanation.                                                                                                                               |
+| `unsafe`      | `(str: string) => TqlTemplateString`                                                                                                 | Accepts a string and returns a representation of the string that will be inserted VERBATIM, UNESCAPED into the compiled query. Please, for all that is good, it's in the name -- this is unsafe. Do not use it unless you absolutely know your input is safe. |
+
+Important: Anywhere you pass a single value into `query` or `fragment`, you can also pass in an array of values. They'll be treated just as if you'd simply interpolated them right next to each other, with all the same protections.
+
+### Parameters
+
+Anything directly passed into a template becomes a parameter. Essentially, the "holes" in the template are filled in with the dialect's parameter placeholder, and the value itself is passed directly into the parameters array:
+
+```ts
+const userId = 1234;
+const [q, params] = query`SELECT * FROM users WHERE user_id = ${userId}`;
+// output: ['SELECT * FROM users WHERE user_id = $1', [1234]]
+```
+
+### List parameters
+
+Need to use list syntax?:
+
+```ts
+const userId = [1234, 5678];
+const [q, params] = query`SELECT * FROM users WHERE user_id IN ${list(userId)}`;
+// output: ['SELECT * FROM users WHERE user_id IN ($1, $2)', [1234, 5678]]
+```
+
+### Composable queries
+
+Need to share clauses between queries, or do you just find it more convenient to build a specific query in multiple variable declarations? `fragment` is what you're looking for!
+
+```ts
+const userId = 1234;
+const whereClause = fragment`WHERE user_id = ${userId}`;
+const [q, params] = query`SELECT * FROM users ${whereClause}`;
+// output: ['SELECT * FROM users WHERE user_id = $1', [1234]]
+```
+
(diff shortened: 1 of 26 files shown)


packages/tql/README.md
```diff
@@ -0,0 +1,166 @@
+# What is it?
+
+```sh
+pnpm i @vercel/tql
+```
+
+TQL (Template-SQL -- unfortunately, T-SQL is already taken!) is a lightweight library for writing SQL in TypeScript:
```
adriancooney: With the `@vercel/` prefix now, you can probably use `tsql` if you want! If not, we can remove this line.

packages/tql/src/build.ts
```diff
@@ -0,0 +1,26 @@
+import { TqlError } from './error';
+import { type TqlQuery, type TqlFragment, type TqlNodeType } from './nodes';
+import type { DialectImpl } from './types';
+
+// TODO: test
```
adriancooney: Is it much effort to add tests here?

packages/tql/src/error.test.ts
```diff
@@ -0,0 +1,27 @@
...
+// eslint-disable-next-line no-new -- this is a typescript test, not a side effect
+new TqlError('illegal_query_recursion', 'abc', 'abc');
+
+// @ts-expect-error - Should error if too few arguments are provided given an error code
+// eslint-disable-next-line no-new -- this is a typescript test, not a side effect
+new TqlError('dialect_method_not_implemented');
+
+// @ts-expect-error - Should error if an argument of the wrong type is provided given an error code
+// eslint-disable-next-line no-new -- this is a typescript test, not a side effect
+new TqlError('dialect_method_not_implemented', 1);
```
adriancooney, line 27: What is happening here? Are we expecting these constructors to throw errors and fail the tests?

packages/tql/src/error.ts
```diff
@@ -0,0 +1,42 @@
+export class TqlError<T extends keyof typeof messages> extends Error {
+  constructor(
+    public code: T,
+    ...details: Parameters<(typeof messages)[T]>
+  ) {
+    // @ts-expect-error - This is super hard to type but it does work correctly
```
adriancooney, line 6: I'm not a big fan of committing these to code unless absolutely necessary but I am acutely aware of how annoying this typing can be. An alternative that simplifies the typing but makes things slightly more verbose could be to create error class for each error type. For example:

```ts
export class TqlError extends Error {}
export class UntemplatedSqlCallTqlError extends TqlError {}
export class DialectMethodNotImplementedTqlError extends TqlError {
  constructor(method: string) {
    super(`The dialect you are using does not implement this method: ${method}`);
  } 
}
```

It also allows for super simple error discrimination for users:

```ts
import { DialiectMethodNotImplementedTqlError } from "@vercel/tql";

try {
  // ..
} catch(err) {
  if (err instanceof DialectMethodNotImplementedTqlError) {
    // .. do work
  }

  throw err;
}
```

packages/tql/src/utils.ts
```diff
@@ -0,0 +1,43 @@
+export function isTemplateStringsArray(
+  value: unknown,
+): value is TemplateStringsArray {
+  return (
+    Array.isArray(value) && Object.prototype.hasOwnProperty.call(value, 'raw')
+  );
+}
+
+// TODO: Test
```
adriancooney: Seems more tests missing here.

packages/tql/README.md
elliott-with-the-longest-name-on-github: T-SQL is still taken (Transact-SQL, Microsoft's dialect used in SQL Server), but I removed the statement now that it's less obvious.

packages/tql/src/error.ts
elliott-with-the-longest-name-on-github: I actually think the ergonomics of subclasses are significantly _worse_ for the user in most cases, though I'm willing to be swayed. The main reason I've gone this route is because it makes this possible:

```ts
catch (e) {
  // as soon as you type the opening quote, TypeScript gives you
  // every possible error from this library, requiring no documentation
  if (e instanceof TqlError && e.code === 'untemplated_sql_call')
}

catch (e) {
  if (e instanceof TqlError) {
    switch (e.type) {
      case 'untemplated_sql_call':
        break;
      default:
        console.log('Impossible if exhaustive');
    }
  }
}
```

as opposed to:

```ts
catch (e) {
  if (e instanceof SomeErrorFromThisLibrary) {
    // I have to both know that this error type exists and import it,
    // for every kind of error I'd like to handle
    // which makes documentation hard and discoverability even harder
  }
}

catch (e) {
  // oops, there's actually no way to switch over error types at all, 
  //I'd have to import every class I want to check for and if/elseif them. 
  //There's no way to make sure I'm handling every error type, 
  // and TypeScript gives me no help in making sure things are exhaustive.
}
```

I'm not particularly worried about this comment, though I'll look into whether there's a way to fix it -- I just couldn't find one in the limited time I had!

packages/tql/src/error.test.ts
elliott-with-the-longest-name-on-github: These are typing tests -- they purely make sure that the types for this error constructor are working correctly. This is how tests are generally constructed when you want to only test typings -- it's how they're set up in `DefinitelyTyped`, for example -- you do something that _should_ cause a type error, then assert that it _is_ causing a type error.

packages/tql/src/dialects/postgres.ts
```diff
@@ -0,0 +1,104 @@
...
+      const rows: string[] = [];
+      for (const entry of entries.values) {
+        validator.validate(entry);
+        if (first) {
+          first = false;
+          columns = Object.keys(entry);
+          this.appendToQuery(
+            `(${columns
+              // eslint-disable-next-line @typescript-eslint/unbound-method -- this rule is idiotic, this is a static method
+              .map(PostgresDialect.escapeIdentifier)
```
adriancooney: I agree this rule is unnecessary but maybe for the sake of clean code we can be a little more verbose?

```suggestion
              .map(column => PostgresDialect.escapeIdentifier(column))
```

(same with down below)

packages/tql/src/error.ts
adriancooney: Ah we can still do best of both worlds! The super-class `TqlError` constructor can require a `code` parameter which the subclass `DialectMethodNotImplementedTqlError` can provide.

```ts
class TqlError extends Error {
	code: string;

	constructor(code: string, message: string) {
		super(message);
		this.code = code;
 	}
}

class DialectMethodNotImplementedTqlError extends TqlError {
  constructor() {
  	super("dialect_method_not_implemented", <message>)
  }
}
```

It's a matter of preference though I think. I generally prefer to discriminate errors using `instanceof`.

packages/tql/src/dialects/postgres.ts
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  list(vals: TqlList): void {
+    this.appendToQuery('(');
+    const queryItems: string[] = [];
+    for (const param of vals.values) {
+      const paramNumber = this.appendToParams(param);
+      queryItems.push(`$${paramNumber}`);
+    }
+    this.appendToQuery(queryItems.join(', '));
```
adriancooney, line 40:
```suggestion
    for (const param of vals.values) {
    	this.parameter(param);
    	this.appendToQuery(", ");
    }
```

Slight simplification. Feel free to ignore.

packages/tql/src/dialects/postgres.ts
```diff
@@ -0,0 +1,104 @@
...
+      this.appendToQuery(
+        // eslint-disable-next-line @typescript-eslint/unbound-method -- this rule is idiotic, this is a static method
+        `(${columns.map(PostgresDialect.escapeIdentifier).join(', ')}) VALUES `,
+      );
+      const queryItems: string[] = [];
+      for (const column of columns) {
+        const paramNumber = this.appendToParams(entry[column]);
+        queryItems.push(`$${paramNumber}`);
+      }
+      this.appendToQuery(`(${queryItems.join(', ')})`);
```
adriancooney, line 84: Can this be replaced with a call to `this.list`?

packages/tql/src/dialects/postgres.ts
```diff
@@ -0,0 +1,104 @@
...
+  type TqlParameter,
+  type TqlTemplateString,
+  type TqlSet,
+  type TqlValues,
+} from '../nodes.js';
+
+export class PostgresDialect extends BaseDialect implements DialectImpl {
+  string(str: TqlTemplateString): void {
+    this.appendToQuery(str.value);
+  }
```
adriancooney: Are we wrapping this value in quotes in the query? If not, is there a reason we call this `string`? It feels like `raw` or something along those lines might be a more apt name.

packages/tql/src/index.ts
```diff
@@ -0,0 +1,84 @@
...
+      const query = parseTemplate(TqlQuery, strings, values);
+      const qb = createQueryBuilder();
+      const d = new Dialect(qb.appendToQuery, qb.appendToParams);
+      const preprocessed = d.preprocess(query);
+      build(d, preprocessed);
+      return d.postprocess(qb.query, qb.params);
+    },
+    fragment: (strings, ...values) =>
+      parseTemplate(TqlFragment, strings, values),
+    identifiers: (ids) => new TqlIdentifiers(ids),
```
adriancooney, line 34: Should `identifiers` be plural here? I know the input can be a list but I think the output is singular. For example, `"myschema"."mytable"` is a singular identifier. Given that, is `identifier` (singular) more appropriate here?

packages/tql/src/index.ts
```diff
@@ -0,0 +1,84 @@
...
+    list: (vals) => new TqlList(vals),
+    values: (entries) => new TqlValues(entries),
+    set: (entries) => new TqlSet(entries),
```
adriancooney, line 37: It's unfortunate that `list` and `set` in this context here are very different than your typical list/set and might confuse users (e.g. `Array` vs `Set`). Is it worth trying to separate the concepts of list/set and list/sql-update-set? Maybe `list` becomes `join` or `array`? (`join` is actually pretty common).

adriancooney (commented):
Overall this is really great. Some small notes about the exported API and missing tests. Great work @tcc-sejohnson!

packages/tql/src/dialects/postgres.ts
elliott-with-the-longest-name-on-github: I'm not sure this would work -- wouldn't it cause there to be a trailing comma at the end of the list?

packages/tql/src/dialects/postgres.ts
elliott-with-the-longest-name-on-github: It's called a `TqlTemplateString` (see the JSDoc on that class) -- though I wouldn't be against renaming it! TemplateStrings are only created by the develper-typed portions of `query` and `fragment` objects -- i.e. in:

```ts
const q = query`SELECT * FROM user WHERE user_id = ${1234} AND LOWER(user_name) LIKE 'jingleheimer%'`
```

You'd get this AST:

```ts
const ast = new TqlQuery([
  new TqlTemplateString('SELECT * FROM user WHERE user_id = '),
  new TqlParameter(1234),
  new TqlTemplateString(" AND LOWER(user_name) LIKE 'jingleheimer%'"),
])
```

Since we prevent `query` and `fragment` from being called in any way other than a template expression, we know these are always safe to append to the query directly, as they only could have been written by the developer. (The one exception is `unsafe`, which creates a `TqlTemplateString` directly, but... well, it's in the name. It's unsafe.)

packages/tql/src/dialects/postgres.ts
elliott-with-the-longest-name-on-github: The point at which ESLint makes clean code less clean to avoid disabling the rule making the clean code error is the point at which ESLint has outlived its usefulness... <picture data-single-emoji=":meow_eyeroll:" title=":meow_eyeroll:">:meow_eyeroll:</picture> 

Fixed, begrudgingly! <picture data-single-emoji=":lol:" title=":lol:">:lol:</picture>

packages/tql/src/dialects/postgres.ts
elliott-with-the-longest-name-on-github: Hmm, you know, that might work in practice, but I think it might grow better if we leave it separate -- I'm not very confident the syntaxes are _always_ the same -- and additionally, we'd have to turn the entry/column map into a `TqlList` before passing it to `this.list`, which probably would end up being more code anyway.

packages/tql/src/index.ts
elliott-with-the-longest-name-on-github: I originally had both `identifier` (accepted a single value) and `identifiers` (accepts multiple values), but I combined them into one -- mirroring the VALUES clause in case, where even if you only have one value it's still VALUES. In this case, it can be called both ways: 

```ts
identifiers(['hello', 'world', "table_a.column_name"]); // output when compiled to Postgres: "hello", "world", "table_a"."column_name"
identifiers('hello'); // output when compiled to Postgres: "hello"
```

I dunno, it seems weird to have two exports just to have the ability to call it with a single value vs. an array, and I'm not sure if making the single export singular or plural is more weird. Thoughts?

packages/tql/src/index.ts
elliott-with-the-longest-name-on-github: I actually hate this just as much as you do! I struggled with `list` a lot and landed on it because:
- It doesn't match the semantics of `join` (you can't specify a delimiter, and it's doing more than just joining, it's actually substituting stuff into your query for you)
- JavaScript doesn't have lists (Python does, but every other programming language I know of calls 'em arrays)
- `array` would be even more confusing (especially since many SQL dialects have some sort of ARRAY or JSON_ARRAY keyword as well)

`set` is, IMO, the most confusing, because... well, JavaScript has a `Set` already. Hmm.

Here's a different idea. The goal with the names of these was to stick as closely as possible to the names of the clauses they output into your SQL. Here's an example -- if I'm writing the following SQL:

```sql
INSERT INTO users ("name", "dob") VALUES ("vercelliott", "2023-01-01")
```

The way that "sounds" when I write it is "INSERT INTO users VALUES". And so, the TQL for it is:

```ts
query`INSERT INTO users ${values(user)}`;
```

It "sounds" the same when you read it.

Here's an idea that might kill two birds with one stone: I have this weird problem with the fact that, just by looking at it, you can't tell which functions of the exported API _create_ queries and which ones go _inside_ queries:

```ts
import {
  query,
  fragment,
  identifiers,
  list,
  values,
  set,
  unsafe, 
} from '@vercel/tql';
```

`query` and `fragment` are "special" in that they're tagged templates, not "normal" functions. All of the other functions are meant to be used "inside" these templates. What if we mirrored the usage of these clauses in SQL, where we typically capitalize key words?:

```ts
import {
  query,
  fragment,
  IDENTIFIERS,
  LIST,
  VALUES,
  SET,
  UNSAFE, 
} from '@vercel/tql';
```

They'd look more at home in the SQL, and you'd also _know_ they're a bit "special" -- they're supposed to be used inside your template:

```ts
query`INSERT INTO users ${VALUES(user)}`;
```

It'd also kinda solve our issue above -- no one will confuse `SET` with `Set`, as things aren't typically capitalized in JavaScript -- but they _are_ in SQL.

elliott-with-the-longest-name-on-github: This works really well in the `SET` case:

```sql
UPDATE users SET name = $1 WHERE user_id = $2;
```

becomes:

```ts
query`UPDATE users ${SET(updatedUser)} WHERE user_id = ${1234}`;
```

which "reads" the same, meaning it'll be easy to remember / write.

packages/tql/src/error.ts
elliott-with-the-longest-name-on-github: Right, but the problem here is that TypeScript is still dumb about them:

```ts
if (tqlError.code === '') // typescript has no idea what these can be
```

and additionally, even if I do know the code of an error:

```ts
if (tqlError.code === 'illegal_query_recursion') {
  // typescript has no way to connect this back to "oh, then the error is actually a `TqlIllegalRecursionError`
}
```

The authorship experience is also somewhat better when using a single error class, as typing:

```ts
throw new TqlError('
```

is all you need to do to see the entire list of errors it's possible to throw from the library, as opposed to having to type `Error` and hope VSCode's intellisense offers to import the right one 😆

packages/tql/src/dialects/postgres.ts
elliott-with-the-longest-name-on-github: Update: I realized there were a bunch of places where the slug for this method was just `string` (probably what you were pointing out), so I updated them all to be `templateString`, which, in addition to the JSDoc on that class, should make things clearer.

packages/tql/src/index.ts
elliott-with-the-longest-name-on-github: I think I may be leaning toward `identifier` which accepts either a single value or array, though it's still kinda weird either way... <picture data-single-emoji=":thinkies:" title=":thinkies:">:thinkies:</picture>

packages/tql/src/index.ts
elliott-with-the-longest-name-on-github: Implemented this in #507 for an example.

vvo:

PR #506, opened 2023-11-28
Spare the comparison and replace with regex

gillyb:
Instead of checking via `.startsWith` and replacing if it is, we can do both these operations via a regular expression replace

### DIFF
--- a/.changeset/loud-llamas-turn.md
+++ b/.changeset/loud-llamas-turn.md
@@ -1,0 +1,5 @@
+---
+'@vercel/postgres': patch
+---
+
+Removed an inline condition to simplify the code just a tiny bit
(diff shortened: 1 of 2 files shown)


.changeset/loud-llamas-turn.md
```diff
@@ -0,0 +1,5 @@
+---
+'@vercel/postgres': patch
+---
+
+Removed an inline condition to simplify the code just a tiny bit
```
dferber90: Not sure I'd even include a changeset for something like this which has no effect for users. 

Having the changeset means this will create a new release, and people will be bothered with upgrading.

dferber90 (approved):

.changeset/loud-llamas-turn.md
gillyb: I didn't think there was a reason to either, but i got the message on github so i thought it was mandatory.  
I will remove it.

dferber90:
Thanks!

PR #512, opened 2023-12-06
[edge-config] Add optional tracing

AndyBitz:
- Exports the `setTracerProvider` to set a trace provider for the Edge Config SDK
 - Adds a tracing wrapper function
 - Wraps the most critical functions to ensure they get traced

Haven't added any tests yet, but confirmed it works by running it with a local setup.

No Changeset because it merges into the `batching-and-deduping` branch.

### DIFF
--- a/packages/edge-config/package.json
+++ b/packages/edge-config/package.json
@@ -49,6 +49,7 @@
     "@changesets/cli": "2.26.2",
     "@edge-runtime/jest-environment": "2.3.6",
     "@edge-runtime/types": "2.2.6",
+    "@opentelemetry/api": "1.7.0",
     "@types/jest": "29.5.7",
     "@types/node": "20.8.10",
     "eslint": "8.52.0",
--- a/packages/edge-config/src/index.ts
+++ b/packages/edge-config/src/index.ts
@@ -128,78 +128,103 @@ export function createClient(
   const loadersInstanceCache = new WeakMap<RequestContext, Loaders>();
 
   const api: Omit<EdgeConfigClient, 'connection'> = {
-    get: trace(async function get<T = EdgeConfigValue | undefined>(
-      key: string,
-    ): Promise<T> {
-      assertIsKey(key);
-      const loaders = getLoadersInstance(loaderOptions, loadersInstanceCache);
+    get: trace(
+      async function get<T = EdgeConfigValue | undefined>(
+        key: string,
+      ): Promise<T> {
+        assertIsKey(key);
+        const loaders = getLoadersInstance(loaderOptions, loadersInstanceCache);
 
-      return loaders.get.load(key).then((value) => {
-        // prime has() with the result of get()
-        loaders.has.prime(key, value !== undefined);
-        return clone(value);
-      }) as Promise<T>;
-    }),
-    async has(key): Promise<boolean> {
-      assertIsKey(key);
-      const loaders = getLoadersInstance(loaderOptions, loadersInstanceCache);
+        return loaders.get.load(key).then((value) => {
+          // prime has() with the result of get()
+          loaders.has.prime(key, value !== undefined);
+          return clone(value);
+        }) as Promise<T>;
+      },
+      {
+        name: 'get',
+      },
+    ),
+    has: trace(
+      async function has(key): Promise<boolean> {
+        assertIsKey(key);
+        const loaders = getLoadersInstance(loaderOptions, loadersInstanceCache);
 
-      return loaders.has.load(key);
-    },
-    async getMany<T = (EdgeConfigValue | undefined)[]>(
-      keys: string[],
-    ): Promise<T> {
-      const loaders = getLoadersInstance(loaderOptions, loadersInstanceCache);
+        return loaders.has.load(key);
+      },
+      {
+        name: 'has',
+      },
+    ),
+    getMany: trace(
+      async function getMany<T = (EdgeConfigValue | undefined)[]>(
+        keys: string[],
+      ): Promise<T> {
+        const loaders = getLoadersInstance(loaderOptions, loadersInstanceCache);
 
-      assertIsKeys(keys);
-      const values = await loaders.get.loadMany(keys);
+        assertIsKeys(keys);
+        const values = await loaders.get.loadMany(keys);
 
-      // throw error in case the edge config could not be found
-      const error = values.find((v): v is Error => v instanceof Error);
-      if (error) throw error;
+        // throw error in case the edge config could not be found
+        const error = values.find((v): v is Error => v instanceof Error);
+        if (error) throw error;
 
-      // prime get() and has() calls with the result of getMany()
-      keys.forEach((key, index) => {
-        if (!key) return;
-        const value = values[index];
-        loaders.get.prime(key, value);
-        loaders.has.prime(key, value !== undefined);
-      });
-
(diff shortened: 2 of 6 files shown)


packages/edge-config/package.json
```diff
@@ -49,6 +49,7 @@
     "@changesets/cli": "2.26.2",
     "@edge-runtime/jest-environment": "2.3.6",
     "@edge-runtime/types": "2.2.6",
+    "@opentelemetry/api": "1.7.0",
```
AAorris, line 52: Did we have a goal to make this a peer dependency?

AndyBitz: I guess that would make more sense? Since we only need the type really 🤔

AAorris (approved):

packages/edge-config/package.json
dferber90: it shouldn't be a peer dep since that would require people to install it and lead to ugly warnings. I think it could potentially be a devDependency if the type gets bundled correctly?

dferber90 (approved):

packages/edge-config/package.json
AndyBitz: Probably, but seems like it doesn't get bundled, at least`index.d.ts` tries to import from `@opentelemetry/api`.

dvoytenko: It should be a peer dependency indeed.

packages/edge-config/src/utils/tracing.ts
```diff
@@ -1,14 +1,82 @@
-import type { TracerProvider, Tracer } from '@opentelemetry/api';
+import type { TracerProvider, Tracer, Attributes } from '@opentelemetry/api';
 import { name as pkgName, version } from '../../package.json';
 
-let tracer: Tracer | undefined;
+let tracer: TracerProvider | undefined;
 
 export function setTracerProvider(tracerProvider: TracerProvider): void {
-  tracer = tracerProvider.getTracer(pkgName, version);
+  tracer = tracerProvider;
```
dvoytenko, line 7: Could you explain this approach? Normally, the tracer is resolved from the `@opentelementy/api`'s `trace.getTracer()`. Are you attempting to keep this strictly type dependency?

packages/edge-config/src/utils/tracing.ts
```diff
@@ -1,14 +1,82 @@
...
-  return tracer;
+function getTracer(): Tracer | undefined {
+  return tracer?.getTracer(pkgName, version);
+}
+
+export function trace<F extends (...args: any) => any>(
+  fn: F,
+  options: {
+    name: string;
+    tags?: Record<string, string | number | boolean>;
```
dvoytenko, line 18: Nit, but `attributes` is a more canonical name.

packages/edge-config/src/utils/tracing.ts
```diff
@@ -1,14 +1,82 @@
...
+    tagSuccess?: (
+      result: ReturnType<F> extends PromiseLike<infer U> ? U : ReturnType<F>,
+    ) => Attributes;
+    tagError?: (error: Error) => Attributes;
+  } = {
+    name: fn.name,
+  },
+): F {
+  const traced = function (this: unknown) {
+    const args = arguments as unknown as unknown[];
```
dvoytenko, line 28: It's better to have this as `function(this: unknown, ...args: unknown[])`.

packages/edge-config/src/utils/tracing.ts
```diff
@@ -1,14 +1,82 @@
...
+    const that = this;
```
dvoytenko, line 29: What's `that` for? Can you just use `this` throughout?

packages/edge-config/src/utils/tracing.ts
```diff
@@ -1,14 +1,82 @@
...
+    const tracer = getTracer();
+    if (!tracer) return fn.apply(that, args);
+
+    return tracer.startActiveSpan(options.name, (span) => {
+      if (options.tags) span.setAttributes(options.tags);
+
+      try {
+        const result = fn.apply(that, args);
+
+        if (result instanceof Promise) {
```
dvoytenko, line 40: Do you have a full control of `result` here? There are still a lot of `PromiseLike` instead of promises. To borrow the code from Next.js to test on whether something is a promise:

```
const isPromise = <T>(p: any): p is Promise<T> => {
  return p !== null && typeof p === 'object' && typeof p.then === 'function'
}
```

packages/edge-config/src/index.ts
```diff
@@ -125,76 +128,103 @@ export function createClient(
...
-    ): Promise<T> {
-      const loaders = getLoadersInstance(loaderOptions, loadersInstanceCache);
+        return loaders.get.load(key).then((value) => {
+          // prime has() with the result of get()
+          loaders.has.prime(key, value !== undefined);
+          return clone(value);
+        }) as Promise<T>;
+      },
+      {
+        name: 'get',
```
dvoytenko, line 145: Does `key` need to be one of the attributes?

packages/edge-config/package.json
dvoytenko: I've been kind of going back and force on this. But I believe this should be a peer dependency in the end. Upon further investigation, a peer dependency doesn't requires a parent to install. The parent only need to install if they want to use `@opentelemetry/api` themselves, which is exactly what we want here.

packages/edge-config/src/utils/tracing.ts
AndyBitz: Re-used what @dferber90 started, but the thought process was that someone who wants to trace the SDK can call `setTracerProvider` with whatever provider they have, potentially just `trace` from `@opentelemetry/api`.

Would it be better to not have `setTracerProvider` and just try to get `trace` from `@opentelemetry/api`?

packages/edge-config/src/utils/tracing.ts
AndyBitz: Yup, was concerned that it might not be consistent with the nested calls, but shouldn't be a problem.

packages/edge-config/src/utils/tracing.ts
AndyBitz: We should, but I guess there never is a guarantee. Will use the same check instead then.

packages/edge-config/src/utils/tracing.ts
dferber90: I'm not too familiar with otel, so using this approach was just based on how I understood the docs

https://opentelemetry.io/docs/concepts/instrumentation/libraries/#getting-a-tracer

image

Totally happy to accept any guidance here

packages/edge-config/src/index.ts
AndyBitz: What would `key` be, and what is it for?

dvoytenko: That's your parameter above :). Sometimes it's convenient to add simple parameters as attributes to ease debugging.

AndyBitz: Ah right, yup, makes sense 👍

packages/edge-config/package.json
grikomsn: +1 for marking as a peer dependency, currently v1.0.0 breaks on edge runtime environments (see #606)

PR #517, opened 2023-12-07
feat(blob): Add multipart option

vvo:
This PR adds a new option to the `put()` method: `multipart: boolean`.

Multipart uploads are useful when uploading a big file to Vercel Blob.
When you use `multipart: true` in a `put()` call. Then:

- We turn the body into a (web) stream.
- We read the stream part by part, buffering into memory only what we need to maximize network utilization.

This current PR is an advanced proof of concept and needs some adjustments.

TODO:
- [x] Test in a browser (might need to adjust import from 'stream')
- [x] Test in an edge function (might need to adjust import form 'stream')
- [x] Add a retry mechanism in fetch when uploading parts
- [x] Feedback and validate API
- [x] Feedback and validate implementation
- [x] Unit tests + errors handling (stream errors) + better handling of cancelling multipart uploads
- [x] Integration tests

Not in scope:
- [ ] configurable retries, there will be a default algorithm
- [ ] configurable concurrent uploads (and memory), there is a default a algorithm


packages/blob/package.json
```diff
@@ -59,7 +59,6 @@
     }
   },
   "dependencies": {
-    "jest-environment-jsdom": "29.7.0",
```
vvo: Not sure why it was here in the first place

packages/blob/src/put-multipart.ts
```diff
@@ -0,0 +1,358 @@
...
+import { Readable } from 'stream';
+import type { BodyInit } from 'undici';
+import { fetch } from 'undici';
+import type { PutBlobApiResponse, PutBlobResult, PutBody } from './put';
+import { getApiUrl, validateBlobApiResponse } from './helpers';
+
+// Most browsers will cap requests at 6 concurrent uploads per domain (Vercel Blob API domain)
+const maxConcurrentUploads = 6;
+
+// 5MB is the minimum part size accepted by Vercel Blob
```
correttojs: can you explain this? wasn't 5Mb the maximum payload size accepted?
do we want to let the user decide the size?

vvo: AWS (and Cloudflare) rules for multipart uploads are:
- each part must be at least 5MB
- only the last part can be less than 5MB

> do we want to let the user decide the size?
Yes I think we should indeed (I said the opposite in the PR but it's an easy addition). There would be two modes:
- multipart: true (automatic mode, we decide of the size of parts etc..)
- multipart: { ... } (manual mode, user provide options).

packages/blob/src/client.ts
```diff
@@ -260,9 +268,18 @@ export async function handleUpload({
...
-      const { pathname, callbackUrl, clientPayload } = body.payload;
-      const payload = await onBeforeGenerateToken(pathname, clientPayload);
+      const { pathname, callbackUrl, clientPayload, multipart } = body.payload;
+      const payload = await onBeforeGenerateToken(
+        pathname,
+        clientPayload,
+        multipart,
+      );
       const tokenPayload = payload.tokenPayload ?? clientPayload;
+      const defaultTokenDurationInSeconds = multipart ? 60 * 60 : 10 * 60;
```
correttojs: Should we always keep a longer default? I think the client (no multipart) upload has more chances to timeout as it takes more time to reach the worker

vvo: Right one hour for everone then?

correttojs: yes

correttojs (approved):

packages/blob/src/client.ts
vvo: Switched to one hour by default.

packages/blob/src/client.ts
```diff
@@ -260,10 +271,21 @@ export async function handleUpload({
...
+      const { pathname, callbackUrl, clientPayload, multipart } = body.payload;
+      const payload = await onBeforeGenerateToken(
+        pathname,
+        clientPayload,
+        multipart,
+      );
       const tokenPayload = payload.tokenPayload ?? clientPayload;
 
+      // one hour
+      const oneHourInSeconds = 60 * 60;
```
luismeyer, line 283: isn't this way too long? I mean it probably doesn't hurt but one hour seems unnecessary long to me

vvo: I would not say way too long, I could see an upload taking 30 minutes already (like the 14gb file I tested), so one hour is just to be safe here

PR #525, opened 2023-12-08
Update Edge Config tracing

AndyBitz:
Updates the Edge Config tracing with feedback from https://github.com/vercel/storage/pull/512


dferber90 (approved):

packages/edge-config/src/utils/tracing.ts
```diff
@@ -1,79 +1,91 @@
...
+  type Tracer,
+  type Attributes,
+} from '@opentelemetry/api';
 import { name as pkgName, version } from '../../package.json';
 
-let tracerProvider: TracerProvider | undefined;
-
-export function setTracerProvider(nextTracerProvider: TracerProvider): void {
-  tracerProvider = nextTracerProvider;
+function getTracer(): Tracer | null {
```
dvoytenko: Nit: this API never returns `null`. But ok if you need it for some reason.

dvoytenko (approved):

packages/edge-config/src/utils/tracing.ts
AndyBitz: Good catch, forgot to remove it from the previous implementation

packages/edge-config/package.json
```diff
@@ -63,6 +62,9 @@
     "tsup": "8.0.1",
     "typescript": "5.3.3"
   },
+  "peerDependencies": {
+    "@opentelemetry/api": "~1.0"
```
dferber90, line 66: This should be a devDependency too, right?

packages/edge-config/src/index.ts
```diff
@@ -66,147 +64,169 @@ interface EdgeConfigClientOptions {
...
+          );
+          if (keys === undefined) {
+            const items = await loaders.getAll.load('#').then(clone);
 
-        assertIsKeys(keys);
-        const values = await loaders.get.loadMany(keys);
+            // prime get() and has() calls with the result of getAll()
+            Object.entries(items).forEach(([key, value]) => {
+              loaders.get.prime(key, value);
+              loaders.has.prime(key, true);
```
dferber90, line 188: In `get` we're priming with `value !== undefined` so I think we should prime the same way here

```suggestion
              loaders.has.prime(key, value !== undefined);
```

packages/edge-config/package.json
AndyBitz: @dferber90 no, we want this to be a peer dependency, because that should allow us to use whatever version of `@opentelemetry/api` the client uses. If we make it a dev dependency, then it might be a different version and might not be able to resolve the `getTracer` call.

packages/edge-config/src/index.ts
AndyBitz: Don't think I've changed that. Maybe better to address in the `batching-and-deduping` branch directly?

PR #540, opened 2023-12-18
lock object

dferber90:
adds lock-object proxy basics


dferber90:
Closing this as similar functionality will be part of our v1 and has found its way to main through other PRs

PR #551, opened 2023-12-27
FIX for bug 550: Kysely migrations fail because VercelPostgresDialect incorrectly reports that transactions are supported

### LINKED ISSUE #550: Kysely Migrations fail because VercelPostgresDialect fails to report that VercelPostgres does not support transactions
The Migrator class in Kysely will use transactions if the provided dialect says that the db supports them:
https://github.com/kysely-org/kysely/blob/680eb46d257f58d437ad8f6334255b48d1e9722d/src/migration/migrator.ts#L425

The VercelPostgresDialect class does provides the standard Kysely PostgresAdapter (it fails to override `PostgresDialect.createAdapter`) which reports that transactions _are_ supported:
https://github.com/kysely-org/kysely/blob/680eb46d257f58d437ad8f6334255b48d1e9722d/src/dialect/postgres/postgres-adapter.ts#L10

This causes the Kysely `Migrator` class to attempt to use transactions when running migrations against Vercel/Neon Postgres, which of course fails.

The solution is for the VercelPostgresDialect to report that transactions are _not_ supported by either:
- Subclassing `PostgresAdapter` to create an adapter where `get supportsTransactionalDdl` returns false and returning it
- Returning a proxy to the regular `PostgresAdapter` that intercepts the `get` and returns false

jwishnie:
Fixes #550 

This patch:
- creates a subclass of `PostgresAdapter` to correctly report that Vercel Postgres does NOT support transactions 
- Updates `VercelPostgesDialect` to returns this new `VercelPostgresAdapter`

As a result, the Kysely `Migrator` class will _not_ use transactions during migrations, and so Kysely migrations work correctly against Vercel Postgres


correttojs:
thanks for the great contribution! Taking over the PR so I'll ensure the tests are passing

jwishnie:
Thank you for the responsiveness! As much as I enjoy `pnmp patch` I'm excited to get rid of my monkey-patched version of the lib!

PR #559, opened 2024-01-11
feat(blob): allow folder creation

luismeyer:


packages/blob/src/put.ts
```diff
@@ -42,23 +42,25 @@ export function createPutMethod<
...
+          | ReadableStream
+          | File,
     options?: T,
   ): Promise<PutBlobResult> {
     if (!pathname) {
       throw new BlobError('pathname is required');
     }
 
-    if (!body) {
+    if (!body && !pathname.endsWith('/')) {
```
vvo: Let's add a comment that says this is the code that makes creating folders work.

vvo: Should we also make it so that if you pass a pathname that ends with '/' and you also pass a body, then we warn about usage?

packages/blob/src/put.ts
```diff
@@ -42,23 +42,25 @@ export function createPutMethod<
...
-      | string
-      | Readable
-      | Blob
-      | ArrayBuffer
-      | FormData
-      | ReadableStream
-      | File,
+  return async function put<TPath extends string>(
+    pathname: TPath,
+    body: TPath extends `${infer _rest}/`
```
vvo: Is this compatible with TS 4 too? Just wondering I never saw it yet. Can you show a screenshot/video of the DX in VSCode?

packages/blob/src/put.ts
luismeyer: we could even error early in here because the API returns an error as well, for all the JS people

packages/blob/src/put.ts
luismeyer: https://github.com/vercel/storage/assets/41523937/0062f813-6110-4154-b67d-185f29b3cb63

luismeyer: i think the `infer` keyword was already part of TS 4 so it should be compatible

luismeyer: also i think this can be done without infer... stay tuned 😅

luismeyer: ok, here is my second shot at this. It's less verbose for the user since no explicit `undefined` is needed but it might be too much magic. 
Also the error is not really helpful when you switch from file to folder and ts thinks the options object is the body

https://github.com/vercel/storage/assets/41523937/22a35098-5c8f-45bc-8134-86a33be80546

vvo: Pretty cool, thank you!

packages/blob/src/put.ts
```diff
@@ -42,26 +42,42 @@ export function createPutMethod<
...
+          | ReadableStream
+          | File,
+    optionsInput?: T,
   ): Promise<PutBlobResult> {
     if (!pathname) {
       throw new BlobError('pathname is required');
     }
 
-    if (!body) {
+    const isEmptyFolder = pathname.endsWith('/');
```
vvo: Should we rather say `const isFolderCreation = pathname.endsWith('/');`?

packages/blob/src/put.ts
```diff
@@ -105,7 +121,7 @@ export function createPutMethod<
 
     const blobApiResponse = await fetch(getApiUrl(`/${pathname}`), {
       method: 'PUT',
-      body: body as BodyInit,
+      body,
```
vvo, line 124: nice!

packages/blob/src/put.ts
vvo: Yes we could do so indeed

vvo (approved):

packages/blob/src/put.ts
luismeyer: done

packages/blob/src/put.ts
vvo: for some reason I had to re-add it myself in the multipart PR, not sure why @luismeyer

PR #563, opened 2024-01-15
feat(blob): allow inline content disposition for certain blobs

### LINKED ISSUE #509: Blob PDF Support
Hi - I'm evaluating Vercel Blob for usage in a project, so I've been uploading and testing with different file types. Images uploaded to Blob render when using img or the NextJS Image component. Video files successfully play in a video tag. However when I upload a PDF to blob, and use iframe, embed or object HTML tags to render, the PDF is downloaded to my computer instead of being viewable inline on my HTML document. Is there a supported way to display PDFs from blob on a HTML page?

luismeyer:
API PR: https://github.com/vercel/edge-functions/pull/1196

fixes https://github.com/vercel/storage/issues/509

### DIFF
--- a/.changeset/cold-hotels-cross.md
+++ b/.changeset/cold-hotels-cross.md
@@ -1,0 +1,13 @@
+---
+"@vercel/blob": minor
+---
+
+feat(blob): allow inline content disposition for certain blobs
+
+Once you use this new version, then most common medias won't be automatically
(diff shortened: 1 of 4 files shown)


vvo:
@luismeyer I added a changelog

vvo (approved):

Issue #487, opened 2023-11-10
[@vercel/blob] – BlobUnknownError and how to catch it?


mfts:
Hi, I'm getting unknown errors from vercel/blob randomly. I cannot pinpoint why that error occurs. 

```
Error: BlobUnknownError: Vercel Blob: Unknown error, please visit https://vercel.com/help
    at validateBlobApiResponse (/Users/mfts/dev/mfts/papermark/node_modules/@vercel/blob/dist/chunk-L5YSJHMQ.cjs:47:13)
    at put (/Users/mfts/dev/mfts/papermark/node_modules/@vercel/blob/dist/chunk-L5YSJHMQ.cjs:140:11)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async __WEBPACK_DEFAULT_EXPORT__ (webpack-internal:///(api)/./pages/api/mupdf/convert-page.ts:46:22)
```

It's happening in this [file](https://github.com/mfts/papermark/blob/main/pages/api/mupdf/convert-page.ts):
- in L27: fetch the blob from blob url and convert to `arrayBuffer`
- in L40: upload a new file to blob 

My suspicion is that the `put` in L40 fails somehow. 

Would appreciate your help to better understand how to prevent/catch this error.

vvo:
@mfts Would you be open to creating a small reproduction repository that would contain everything for us to run it? That would be awesome.

mfts:
@vvo I'll try to get to it over the weekend. But it's not a deterministic error. Sometimes the same file doesn't work the first time and then works the second time or vice versa.

luismeyer:
hey @mfts,
it's been a while since you opened this issue. Are you still seeing random errors on `put` requests? 
We just released `@vercel/blob@0.19.0` which adds automatic retries for all blob API requests. This could solve your issue, because it was likely caused by internal request problems to our upstream provider.

It would be great if you could let us know if the issue still persists

mfts:
Sorry for sleeping on this issue. I have actually not experienced this issue recently. Thanks for following up with a new blob release. 

closing this issue

PR #566, opened 2024-01-15
feat(blob): add retry to all API requests

luismeyer:
This PR:
- moves all API related stuff into a new file
- consolidates all API fetches into a `requestAPI` function
- this `requestAPI` function takes care of api-version, authorization, response validation, error handling
- also the `requestAPI` function uses `async-retry` to retry failing requests


luismeyer:
probably fixes https://github.com/vercel/storage/issues/487

packages/blob/src/del.ts
```diff
@@ -20,17 +12,13 @@ export async function del(
...
-    method: 'POST',
-    headers: {
-      ...getApiVersionHeader(),
-      authorization: `Bearer ${getTokenFromOptionsOrEnv(options)}`,
-      'content-type': 'application/json',
+  await requestApi(
+    '/delete',
+    {
+      method: 'POST',
+      headers: { 'content-type': 'application/json' },
```
vvo, line 19: I guess we could pass content-type: 'application/json' all the time without any issues or problem, thoughts?

packages/blob/src/helpers.ts
```diff
@@ -1,14 +1,17 @@
...
   /**
    * Define your blob API token.
    * @defaultvalue process.env.BLOB_READ_WRITE_TOKEN
    */
   token?: string;
+  /**
+   * The amount of times to retry the API requests if they fail.
+   * @defaultvalue 10
+   */
+  retries?: number;
```
vvo: I wonder if this is worth exposing to customers. For the first version, I would not expose it just yet and see when someone wants it. As this would give us context.

luismeyer: i also exposed it so i can disable retries for some negative tests that should directly fail

packages/blob/src/api.ts
```diff
@@ -0,0 +1,160 @@
...
+    }
+  }
+}
+
+export async function requestApi<TResponse>(
+  pathname: string,
+  init: RequestInit,
+  commandOptions: BlobCommandOptions | undefined,
+): Promise<TResponse> {
+  init.headers = {
```
vvo: I know this is not checked by eslint, but can we avoid mutating arguments? And instead compute new requestInit base on the function `init`.

```
const newInit = {...}
```

packages/blob/src/helpers.ts
vvo: I see you may have exposed it for testing purposes? If that's the case let's make it so the unit test replies 10 times with a throw.

Why I don't want to expose it yet: it's easier to add than remove in APIs, so until we're sure there's a need to change the number of retries, let's not expose it.

luismeyer: the problem wasn't that it threw 10 times. The test just always timed out. I thought about other ways to control the retry from outside like env vars, but i think thats more dangerous than having this explicit option

vvo: After 10 retries, if the mocked api fails 10 times, send it will throw the correct error as before?

packages/blob/src/index.node.test.ts
```diff
@@ -148,9 +150,9 @@ describe('blob client', () => {
         })
         .reply(502, { error: { code: 'service_unavailable' } });
 
-      await expect(head(`${BLOB_STORE_BASE_URL}/foo-id.txt`)).rejects.toThrow(
-        new BlobServiceNotAvailable(),
-      );
+      await expect(
+        head(`${BLOB_STORE_BASE_URL}/foo-id.txt`, { retries: 0 }),
```
vvo: Here, if you want to test that, at some point we will receive `BlobServiceNotAvailable`, then there should be a way to have the Mock consistently fail, unless I am missing something?

packages/blob/src/helpers.ts
luismeyer: yes. I guess we could make the test long running but than the CI is unnecessary long

vvo: Ohh right because the retries are time based, can we advance time with Jest? Since this is a unit test, it could work fine.

Otherwise:
> but i think thats more dangerous than having this explicit option

We do this with `VERCEL_BLOB_API_URL` and it's working fine, perhaps we can still go this route?

luismeyer: it doesn't work with the fake timers, so I added `VERCEL_BLOB_RETRIES`

packages/blob/src/del.ts
luismeyer: i think actually only `del` and `multipart-complete` **need** `application/json`

vvo (approved):
LGTM, tests are passing, thanks!

packages/blob/src/api.ts
```diff
@@ -0,0 +1,171 @@
...
+          authorization: `Bearer ${getTokenFromOptionsOrEnv(commandOptions)}`,
+
+          ...init.headers,
+        },
+      });
+
+      if (res.status >= 500) {
+        // this will be retried and shown to the user if no more retries are left
+        await validateBlobApiResponse(res);
+      }
```
correttojs, line 159: shouldn't we bail on known errors? all except `unknown_error` and `service_unavailable`, no?

luismeyer: true

correttojs: the [async-retry](https://www.npmjs.com/package/async-retry) package offers a `bail` function to be able to throw an error without retrying it. I think we need something like
```
if (code === 'unknown_error' || code === 'service_unavailable') {
   // retry
   throw getBlobError(error);
} if (code) {
  // do NOT retry
  bail(getBlobError(error))
}
```
otherwise we'd return a response containing the error

correttojs: it would be nice to have a unit test for this code, as I'm not completely sure how it behaves

luismeyer: ah thats nice. It work currently because i check the response outside of the retry again. But I can remove that than

luismeyer: added some tests

correttojs (approved):
great job!

PR #570, opened 2024-01-17
[edge-config] Use `privateEdgeConfig` when available

AndyBitz:
When the `privateEdgeConfig` is available in the global scope, the SDK will try to call its `get` function to read the Edge Config.

The value returned by `privateEdgeConfig.get` is expected to be readonly and throw on any write attempt, even on nested properties. To help mitigate this, we're also exposing a `clone` function which allows to copy a value and make it writeable. To support this behaviour, Edge Config will now wrap all values in `DeepReadonly<T>`.

We do this to save CPU time on parsing and cloning.

---

There is no Changeset as it'll be part of a larger 1.0.0 release.


packages/edge-config/README.md
```diff
@@ -77,6 +77,12 @@ The `createClient` function connects to a any Edge Config based on the provided
 
 It returns the same `get`, `getAll`, `has` and `digest` functions as the default Edge Config Client exports.
 
+### Making a value mutable
+
+By default, the value returned by `get` and `getAll` are immutable. Modifying the object might cause an error or other undefined behaviour.
+
+In order to make the returned value mutable, you can use the exported function `cloneEdgeConfig` to safely clone the object and make it mutable.
```
dferber90:
```suggestion
In order to make the returned value mutable, you can use the exported function `clone` to safely clone the object and make it mutable.
```

dferber90 (changes requested):

packages/edge-config/README.md
```diff
@@ -77,6 +77,12 @@ The `createClient` function connects to a any Edge Config based on the provided
 
 It returns the same `get`, `getAll`, `has` and `digest` functions as the default Edge Config Client exports.
 
+### Making a value mutable
+
+By default, the value returned by `get` and `getAll` are immutable. Modifying the object might cause an error or other undefined behaviour.
```
dferber90:
```suggestion
By default, the value returned by `get` and `getAll` is immutable. Modifying the object might cause an error or other undefined behaviour.
```

packages/edge-config/src/index.ts
```diff
@@ -81,6 +83,49 @@ const getFileSystemEdgeConfig = trace(
...
 );
 
+/**
+ *
+ */
+const getPrivateEdgeConfig = trace(
+  async function getPrivateEdgeConfig(
+    connection: Connection,
+  ): Promise<DeepReadonly<EmbeddedEdgeConfig> | null> {
+    const privateEdgeConfig = Reflect.get(
```
dferber90, line 93: What's the advantage of `Reflect.get` here over `globalThis[privateEdgeConfigSymbol]`?

dferber90 (approved):

packages/edge-config/src/index.ts
AndyBitz: @dferber90 no need to cast `globalThis`, just the returned type.

PR #573, opened 2024-01-22
[edge-config] Fix SWR in Edge Config development cache

AndyBitz:
Fixes a bug in the SWR functionality for Edge Config that is used during development.

At the moment, we don't consider the argument, causing the e.g. `get` function to always return the most recent promise. With this fix, we'll make sure to only return the value if the arguments of the function match.

No Changeset to make it part of the 1.0.0 release.

### DIFF
--- a/packages/edge-config/src/utils/swr-fn.ts
+++ b/packages/edge-config/src/utils/swr-fn.ts
@@ -8,10 +8,12 @@ import { clone } from './index';
 export const swr = trace(
   function swr<T extends (...args: any[]) => Promise<any>>(fn: T): T {
     let latestInvocationId = 0;
-    let staleValuePromise: null | Promise<unknown> = null;
+    const staleValuePromiseMap = new Map<string, Promise<unknown>>();
 
     return (async (...args: any[]) => {
+      const argsKey = JSON.stringify(args);
       const currentInvocationId = ++latestInvocationId;
+      const staleValuePromise = staleValuePromiseMap.get(argsKey);
 
       if (staleValuePromise) {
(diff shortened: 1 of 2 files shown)


packages/edge-config/src/utils/swr-fn.ts
```diff
@@ -8,18 +8,20 @@ import { clone } from './index';
 export const swr = trace(
   function swr<T extends (...args: any[]) => Promise<any>>(fn: T): T {
     let latestInvocationId = 0;
-    let staleValuePromise: null | Promise<unknown> = null;
+    const staleValuePromiseMap = new Map<string, Promise<unknown>>();
 
     return (async (...args: any[]) => {
+      const argsKey = JSON.stringify(args);
       const currentInvocationId = ++latestInvocationId;
+      const staleValuePromise = staleValuePromiseMap.get(argsKey);
```
dferber90, line 14: Would the `currentInvocationId` and `latestInvocationId` need to be per argsKey? Otherwise the cache might not get set later on

dferber90: I think we could refactor this whole thing to get rid of the `swr` function. Instead of fetching individual items, we want to fetch the whole edge config in dev mode. We could add this to `index.ts` directly, since it already deals with "get the right info out of all edge config items when a function like get is called"

AndyBitz: Good point. I removed the counter to just compare the promise. We should only overwrite it, if it's still the same

AndyBitz:
Closing in favour of https://github.com/vercel/storage/pull/574

PR #574, opened 2024-01-22
[edge-config] Fix SWR during development

AndyBitz:
Fixes a bug in the SWR functionality for Edge Config that is used during development.

At the moment, we don't consider the argument, causing the e.g. get function to always return the most recent promise. With this fix, we'll keep the entire Edge Config in memory and revalidate it only on a request in the background.

No Changeset to make it part of the 1.0.0 release.


packages/edge-config/src/index.common.test.ts
```diff
@@ -389,3 +389,32 @@ describe('connectionStrings', () => {
...
+    const edgeConfig = pkg.createClient(
+      'https://edge-config.vercel.com/ecfg-2?token=token-2',
+    );
+    expect(await edgeConfig.get('foo')).toBe('bar');
+
+    fetchMock.mockResponse(JSON.stringify({ foo: 'bar2' }));
+    expect(await edgeConfig.get('foo')).toBe('bar'); // 1st call goes to the cache
+
+    await new Promise<void>((res) => {
+      setTimeout(res, 1000);
```
dferber90: Does this really need to be 1s? This would slow the tests down right?

packages/edge-config/src/index.ts
```diff
@@ -113,6 +112,70 @@ const getPrivateEdgeConfig = trace(
   },
 );
 
+/**
+ *
```
dferber90: Description isn't saying a lot at the moment <picture data-single-emoji=":jjj_lol:" title=":jjj_lol:">:jjj_lol:</picture>

packages/edge-config/src/index.ts
```diff
@@ -113,6 +112,70 @@ const getPrivateEdgeConfig = trace(
...
+ */
+function createGetInMemoryEdgeConfig(
+  useDevelopmentCache: boolean,
+  connection: Connection,
+  headers: Record<string, string>,
+): () => Promise<EmbeddedEdgeConfig | null> {
+  let embeddedEdgeConfigPromise: Promise<EmbeddedEdgeConfig | null> | null =
```
dferber90, line 127: A description of embeddedEdgeConfigPromise and lastRequest would help make the code below more readable. At the moment I get it roughly but it's hard to follow some of the nuance

packages/edge-config/src/index.ts
```diff
@@ -113,6 +112,70 @@ const getPrivateEdgeConfig = trace(
...
+    null;
+  let lastRequest: Promise<EmbeddedEdgeConfig | null> | null = null;
```
dferber90: Should this be `latestRequest`?

packages/edge-config/src/index.ts
```diff
@@ -208,17 +271,25 @@ export const createClient = trace(
      * While in development we use SWR-like behavior for the api client to
      * reduce latency.
      */
-    const shouldUseSwr =
+    const useDevelopmentCache =
```
dferber90: We shouldn't prefix with "use" due to confusion with react hooks. Not sure it would actually lead to errors, but better to avoid this in the first place

```suggestion
    const shouldUseDevelopmentCache =
```

dferber90 (changes requested):
So much simpler <picture data-single-emoji=":frog-10:" title=":frog-10:">:frog-10:</picture>

packages/edge-config/src/index.common.test.ts
AndyBitz: Not really, reduced it to 100ms.

It works even without any waiting, but less predictable, since we can't know when the promise gets resolved.

dferber90 (approved):

PR #577, opened 2024-01-23
[edge-config] Changeset for major release

AndyBitz:
### Release Edge Config 1.0.0

See the attached Changeset for the relevant changes.


.changeset/lemon-phones-sort.md
```diff
@@ -0,0 +1,9 @@
+---
+"@vercel/edge-config": major
+---
+
+- Use stale-while-revalidate during development
+- Don't throw generic errors
+- Make the return values read only
+- Use the `privateEdgeConfig` global when available
+- Export a clone function to support the read only behaviour
```
dferber90: Let's add an upgrade guide for people who are upgrading from our current "latest" release.

What are the things they need to do to be compatible with the v1 release?

To my mind comes: If somebody was modifying the return values, they now need to explicitly call `clone`. We should have an example for that in here

.changeset/lemon-phones-sort.md
```diff
@@ -0,0 +1,9 @@
+---
+"@vercel/edge-config": major
+---
+
+- Use stale-while-revalidate during development
+- Don't throw generic errors
+- Make the return values read only
+- Use the `privateEdgeConfig` global when available
```
dferber90: Let's not list this as it's an internal change which is not relevant for users

.changeset/lemon-phones-sort.md
```diff
@@ -0,0 +1,9 @@
+---
+"@vercel/edge-config": major
+---
+
+- Use stale-while-revalidate during development
```
dferber90: Let's highlight a bit more why we're doing this as it's quite nice

.changeset/lemon-phones-sort.md
```diff
@@ -0,0 +1,9 @@
...
+- Don't throw generic errors
```
dferber90: Let's detail that applications now need to handle more than just the network error.

dferber90 (changes requested):

.changeset/lemon-phones-sort.md
```diff
@@ -2,8 +2,21 @@
 "@vercel/edge-config": major
 ---
 
-- Use stale-while-revalidate during development
-- Don't throw generic errors
-- Make the return values read only
-- Use the `privateEdgeConfig` global when available
-- Export a clone function to support the read only behaviour
+ - **Return values are now read-only to improve in-memory caching**
+   In addition to the read-only change, the `@vercel/edge-config` package now also exports a `clone` function that can be used to make the read-only values modifiable again. Applications that used to modify the return value must now clone the value before doing the required changes.
```
dferber90: An intro paragraph with an example would be good. then the "in addition to.." is anchor in something more specific.

> Before this change it was possible to modify the values returned by the Edge Config SDK's methods. For example this was possible
>
> ```
> const someValue = await get("someKey")
> someValue.foo = 1
> ```
> Going forward any modifications will result in an error as the values returned by the SDK are now read-only.

.changeset/lemon-phones-sort.md
```diff
@@ -2,8 +2,21 @@
...
+
+   ```typescript
+    import { get, clone } from '@vercel/edge-config';
+
+    const myArray = await get('listOfAllowedIPs');
+    // myArray.push('127.0.0.1'); // This won't work and might throw an error
```
dferber90:
```suggestion
    // myArray.push('127.0.0.1'); // This won't work and will throw an error
```

.changeset/lemon-phones-sort.md
```diff
@@ -2,8 +2,21 @@
...
+    const myArrayClone = clone(myArray); // Clones the data to make it modifiable
+    myArrayClone.push('127.0.0.1');
+   ```
```
dferber90, line 25: Another paragraph in the end like

> Cloning the return value is only necessary in case the values will be modified.

.changeset/lemon-phones-sort.md
```diff
@@ -2,8 +2,21 @@
...
+
+ - **Use stale-while-revalidate during development**
+   When `@vercel/edge-config` is used during development, with `NODE_ENV` being set to `development`, the most recent result will be kept in-memory and returned instead of waiting for another network request to complete. In the background, a new request will then be started to ensure the in-memory data gets updated.
```
dferber90: This could be a bit more precise, since we are not keeping the last result in memory, but we are actually fetching the full edge config no matter which key is read and then use that to respond to further reads, while refreshing in the background.

.changeset/lemon-phones-sort.md
```diff
@@ -2,8 +2,21 @@
...
+   This behaviour can be changed by setting the environment variable `EDGE_CONFIG_DISABLE_DEVELOPMENT_SWR` to `1`, or by using the `disableDevelopmentCache` option on the `createClient` function.
+
+ - **Genric catch-all errors are not thrown anymore**
```
dferber90:
```suggestion
 - **BREAKING CHANGE** SDK now throws underlying errors
 Previous versions of the `@vercel/edge-config` package would catch most errors thrown by native functions and throw a generic network error instead - even if the underlying issue wasn't a network error. The new version will throw the original errors.
 
 **Note** applications which rely on the `@vercel/edge-config: Unexpected error` and `@vercel/edge-config: Network error` errors must adapt to the new implementation by ensuring other types of errors are handled as well.
```

.changeset/lemon-phones-sort.md
```diff
@@ -2,8 +2,21 @@
 "@vercel/edge-config": major
 ---
 
-- Use stale-while-revalidate during development
-- Don't throw generic errors
-- Make the return values read only
-- Use the `privateEdgeConfig` global when available
-- Export a clone function to support the read only behaviour
+ - **Return values are now read-only to improve in-memory caching**
```
dferber90:
```suggestion
 - **BREAKING CHANGE** Return values are now read-only to improve in-memory caching
```

.changeset/lemon-phones-sort.md
```diff
@@ -2,8 +2,21 @@
...
+   ```typescript
+    import { get, clone } from '@vercel/edge-config';
+
+    const myArray = await get('listOfAllowedIPs');
+    // myArray.push('127.0.0.1'); // This won't work and might throw an error
+    const myArrayClone = clone(myArray); // Clones the data to make it modifiable
+    myArrayClone.push('127.0.0.1');
+   ```
+
+ - **Use stale-while-revalidate during development**
```
dferber90:
```suggestion
 - SDK now uses stale-while-revalidate semantics during development
```

.changeset/lemon-phones-sort.md
AndyBitz: It'll cause a TypeScript error a build time, but at runtime it'd only throw in certain cases when the object is truly not writeable beyond just the TypeScript type.

dferber90: Can our proxy object catch this and ensure we throw?

AndyBitz: @dferber90 it does, the reason why it only "might" throw is because the global that creates the Proxy object is not always present, only in Serverless Functions.

AAorris (approved):

.changeset/lemon-phones-sort.md
```diff
@@ -0,0 +1,37 @@
...
+---
+
+ - **BREAKING CHANGE** Return values are now read-only to improve in-memory caching
+
+   It used to be possible to change the returned value as seen this example:
+
+   ```typescript
+    import { get } from '@vercel/edge-config';
+    const countries = await get('allowedCountryCodes');
+    countries.DE = true; // Won't be possible anymore and throw an error
```
dferber90: This will not actually throw anymore now that we removed the proxy, right?

dferber90 (changes requested):
the changelog reads great but there seems to be a factual error - unless we actually kept the proxy which throws?

.changeset/lemon-phones-sort.md
AndyBitz: Yes, updated it now

.changeset/lemon-phones-sort.md
```diff
@@ -0,0 +1,37 @@
+---
+"@vercel/edge-config": major
+---
+
+ - **BREAKING CHANGE** Return values are now read-only to improve in-memory caching
+
+   It used to be possible to change the returned value as seen this example:
```
dferber90:
```suggestion
   It used to be possible to change the returned value as shown in this example:
```

.changeset/lemon-phones-sort.md
```diff
@@ -0,0 +1,37 @@
...
+
+   ```typescript
+    import { get } from '@vercel/edge-config';
+    const countries = await get('allowedCountryCodes');
+    countries.DE = true; // Will now cause TypeScript to error
+   ```
+
+   Moving forward, modifications like the above will cause an error.
```
dferber90:
```suggestion
   Moving forward, modifications like the above will cause a type error.
```

dferber90 (approved):

PR #582, opened 2024-01-25
feat(blob): add manual multipart upload methods

luismeyer:


packages/blob/src/put.ts
```diff
@@ -1,60 +1,28 @@
...
-  getToken?: (pathname: string, options: T) => Promise<string>;
-  extraChecks?: (options: T) => void;
-}) {
+}: CreatePutMethodOptions<TOptions>) {
   return async function put<TPath extends string>(
     pathname: TPath,
-    bodyOrOptions: TPath extends `${string}/` ? T : PutBody,
-    optionsInput?: T,
+    bodyOrOptions: TPath extends `${string}/` ? TOptions : PutBody,
+    optionsInput?: TPath extends `${string}/` ? never : TOptions,
```
luismeyer, line 24: small fix to prevent accepting options input twice

test/next/public/big-text.txt
luismeyer, line 1: added this for testing. It's used in the `script.mts`

packages/blob/src/client.ts
```diff
@@ -104,18 +151,93 @@ export const upload = createPutMethod<UploadOptions>({
...
+type ClientMultipartUploadCommandOptions = ClientCommonCreateBlobOptions &
+  CommonMultipartPutOptions &
+  CommonUploadOptions;
+
+export const multipartUpload =
+  createMultipartPutMethod<ClientMultipartUploadCommandOptions>({
+    allowedOptions: ['contentType'],
+    extraChecks: createUploadExtraChecks('client/`multipartUpload`'),
+    async getToken(pathname, options) {
+      return retrieveClientToken({
```
luismeyer: this will fetch a new client token for each part. I'm not sure if we should do it like this or if we should only request a client token once in `createMultipartUpload` and reuse it for `upload` and `complete`

.changeset/ninety-ladybugs-itch.md
```diff
@@ -0,0 +1,56 @@
+---
+"@vercel/blob": minor
+"vercel-storage-integration-test-suite": patch
+---
+
+# feat(blob): add manual multipart upload methods
```
vvo:
```suggestion
# feat(blob): add advanced multipart upload methods
```

.changeset/ninety-ladybugs-itch.md
```diff
@@ -0,0 +1,56 @@
...
+
+This exposes the three different multipart steps as functions of the SDK. Before this change every multipart upload was uncontrolled, meaning the full data was passed to the SDK and it took care of chunking and uploading.
+Now it's possible to manually upload chunks and start and complete the multipart upload. All of the new functions can be used both on the server and the browser. There are two different API's that can be used.
+
+> When using manual multipart uploads **every upload part has to be of the same size** and **the callside has to take care of memory management and concurrent uploads**!
```
vvo:
```suggestion
```

.changeset/ninety-ladybugs-itch.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Fully manual
```
vvo:
```suggestion
## Individual methods
```

.changeset/ninety-ladybugs-itch.md
```diff
@@ -0,0 +1,56 @@
...
+
+Use `createMultipartPut`, `multipartPut` and `completeMultipartPut` to manage the upload.
```
vvo: Can we rename the individual methods to:
`createMultipartUpload`, `uploadPart`, `completeMultipartUpload` instead? This would more closely map S3 in this case and differentiate from "put"

.changeset/ninety-ladybugs-itch.md
```diff
@@ -0,0 +1,56 @@
...
+  uploadId,
+});
+```
+
+## Semi manual
+
+Since there is some data that stays the same across all multipart methods you can also use the util function that are returned by `createMultipartPut`. We keep some data in the closure of `createMultipartPut` and therefor expose a simpler `put` and `complete` function.
+
+```ts
+const multiPartUpload = await vercelBlob.createMultipartPut('big-file.txt', {
```
vvo: So the individual method createMultipartPut can also be used for the stateful version?

.changeset/ninety-ladybugs-itch.md
```diff
@@ -0,0 +1,56 @@
...
+);
+
+const blob = await vercelBlob.completeMultipartPut(fullPath, [part1, part2], {
+  access: 'public',
+  key,
+  uploadId,
+});
+```
+
+## Semi manual
```
vvo:
```suggestion
## Stateful multipart uploads
```

.changeset/ninety-ladybugs-itch.md
```diff
@@ -0,0 +1,56 @@
...
+
+Since there is some data that stays the same across all multipart methods you can also use the util function that are returned by `createMultipartPut`. We keep some data in the closure of `createMultipartPut` and therefor expose a simpler `put` and `complete` function.
+
+```ts
+const multiPartUpload = await vercelBlob.createMultipartPut('big-file.txt', {
+  access: 'public',
+});
+
+const part1 = await multiPartUpload.put(1, createReadStream(fullPath));
```
vvo: Can we rename put here to `uploadPart`?

.changeset/ninety-ladybugs-itch.md
vvo: Reasoning: the feature we market is "multipart uploads", not "multipart put" 👍

vvo: And yes it could be a little confusing with the generic `upload` method of client uploads but that's ok

luismeyer: I used the names `createMultipartUpload`, `uploadPart`, `completeMultipartUpload` for the client methods that automatically fetch the client token, just like we have it for `put`. How would you name these?

luismeyer: I don't really see the benefit of beeing close to the S3 API, since we want to provide a better DX. I think it doesn't make that much of a difference if we call it `upload` or `put`, it just feels more consistent within our SDK

.changeset/ninety-ladybugs-itch.md
luismeyer: it's not really that `stateful` i suppose. It just keeps `pathname`, `key` and `uploadId` in the closure to expose a simpler API

.changeset/ninety-ladybugs-itch.md
luismeyer: is the suggested change to remove this? Why?

.changeset/ninety-ladybugs-itch.md
luismeyer: yep `createMultipartPut` is the entry for both. It returns `key` and `uploadId` but also the utils

.changeset/ninety-ladybugs-itch.md
vvo: Let me add a suggestion below

.changeset/ninety-ladybugs-itch.md
```diff
@@ -0,0 +1,56 @@
+---
+"@vercel/blob": minor
+"vercel-storage-integration-test-suite": patch
+---
+
+# feat(blob): add advanced multipart upload methods
+
+This exposes the three different multipart steps as functions of the SDK. Before this change every multipart upload was uncontrolled, meaning the full data was passed to the SDK and it took care of chunking and uploading.
```
vvo:
```suggestion
This exposes the three different multipart steps as functions of the SDK. Before this change every multipart upload was uncontrolled, meaning the full data was passed to the SDK and the SDK took care of chunking and uploading.
```

vvo:
@luismeyer I updated the changelog, can you check this is correct? c7b7535b3f927e6214e9695afcc194bd4f8f74e2

I do not think there should be a `multipartUpload` method. The only new methods are:
- createMultipartUpload
- uploadPart
- completeMultipartUpload
- createMultipartUploader

I believe you may have named `uploadPart` `multipartUpload`?  If so let's change that and then we'll be good.

luismeyer:
> @luismeyer I updated the changelog, can you check this is correct? [c7b7535](https://github.com/vercel/storage/commit/c7b7535b3f927e6214e9695afcc194bd4f8f74e2)
> 
> I do not think there should be a `multipartUpload` method. The only new methods are:
> 
> * createMultipartUpload
> * uploadPart
> * completeMultipartUpload
> * createMultipartUploader
> 
> I believe you may have named `uploadPart` `multipartUpload`? If so let's change that and then we'll be good.

oh right. Renamed the method

packages/blob/src/client.ts
```diff
@@ -54,15 +66,67 @@ export const put = createPutMethod<ClientPutCommandOptions>({
...
+
+export const put = createPutMethod<ClientPutCommandOptions>({
+  allowedOptions: ['contentType'],
+  extraChecks: createPutExtraChecks('client/`put`'),
 });
 
-// client.upload()
-// This is a client-side wrapper that will fetch the client token for you and then upload the file
-export interface UploadOptions extends CommonCreateBlobCommandOptions {
+// vercelBlob.multipartUpload()
```
```suggestion
// vercelBlob. createMultipartUpload()
// vercelBlob. uploadPart()
// vercelBlob. completeMultipartUpload()
// vercelBlob. createMultipartUploaded()
```

packages/blob/src/index.ts
```diff
@@ -55,3 +63,28 @@ export { list } from './list';
 
 export type { CopyBlobResult, CopyCommandOptions } from './copy';
 export { copy } from './copy';
+
+// vercelBlob.multipartUpload()
```
vvo:
```suggestion
// vercelBlob. createMultipartUpload()
// vercelBlob. uploadPart()
// vercelBlob. completeMultipartUpload()
// vercelBlob. createMultipartUploaded()
```

vvo (approved):

PR #583, opened 2024-01-26
feat(blob): add blob download url

luismeyer:
needs https://github.com/vercel/infra/pull/6667 and https://github.com/vercel/edge-functions/pull/1249

### DIFF
--- a/packages/blob/src/index.ts
+++ b/packages/blob/src/index.ts
@@ -2,7 +2,7 @@ import type { PutCommandOptions } from './put';
 import { createPutMethod } from './put';
 
 // expose generic BlobError and download url util
-export { BlobError, getDownloadURL } from './helpers';
+export { BlobError, getDownloadUrl } from './helpers';
 
 // expose api BlobErrors
 export {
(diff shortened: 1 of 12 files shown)


vvo (commented):
Does this PR already includes changes from our API where downloadUrl is sent as the blob response? I believe we need to upgrade the API_VERSION too?

luismeyer:
> Does this PR already includes changes from our API where downloadUrl is sent as the blob response? I believe we need to upgrade the API_VERSION too?

why do we need a new API_VERSION. Technically it works for all versions. For older blobs the url and downloadUrl will behave the same but that's ok isn't it?

packages/blob/src/index.ts
```diff
@@ -1,8 +1,8 @@
 import type { PutCommandOptions } from './put';
 import { createPutMethod } from './put';
 
-// expose generic BlobError
-export { BlobError } from './helpers';
+// expose generic BlobError and download url util
+export { BlobError, getDownloadURL } from './helpers';
```
vvo: Let's never uppercase fully URL, let's only use camelCaseUrl everywhere (public and private APIs)

luismeyer: done

vvo:
@luismeyer so we do not forget:
- [x] document getDownloadUrl in CHANGELOG
- [ ] document getDownloadUrl and downloadUrl on website

vvo (approved):

PR #584, opened 2024-01-29
refactor(blob): avoid exposing api responses directly

luismeyer:
This will help is in the future when doing non breaking changes. From now on we don't need to increase our internal API_VERSION for changes like adding a field to the API. This will avoid the automatic exposure of new fields and require a minor version of the npm package

### DIFF
--- a/.changeset/dull-gorillas-unite.md
+++ b/.changeset/dull-gorillas-unite.md
@@ -1,8 +1,8 @@
 ---
 "@vercel/blob": minor
-"vercel-storage-integration-test-suite": patch
+"vercel-storage-integration-test-suite": minor
 ---
 
-# feat(blob): add blob download url
+# feat(blob): add `downloadUrl` and `getDownloadUrl`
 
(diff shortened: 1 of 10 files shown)


.changeset/dull-gorillas-unite.md
```diff
@@ -1,5 +1,5 @@
 ---
-"@vercel/blob": patch
+"@vercel/blob": minor
```
vvo: Let's avoid modifying the changelog (but I guess this change will be ignored if we merge https://github.com/vercel/storage/pull/586)

packages/blob/src/client.browser.test.ts
```diff
@@ -20,6 +20,7 @@ jest.mock('undici', () => ({
       json: () =>
         Promise.resolve({
           url: `https://storeId.public.blob.vercel-storage.com/superfoo.txt`,
+          downloadUrl: `https://storeId.public.blob.vercel-storage.com/superfoo.txt?download=1`,
```
vvo, line 23: I wonder why the refactor requires to update the tests? I thought the feature would have required the test to be updated already

vvo (approved):

packages/blob/src/client.browser.test.ts
luismeyer: because it resolved to `downloadUrl: undefined` now

PR #589, opened 2024-02-05
fix(blob): remove multipart boolean from copy options

luismeyer:


vvo (approved):
As long as we check that this has no impact on other methods (i.e. multipart is still an option on put() and upload() server and client), we're good 👍

luismeyer:
> As long as we check that this has no impact on other methods (i.e. multipart is still an option on put() and upload() server and client), we're good 👍

yep still works on put and upload. The client file has it's own typing

PR #592, opened 2024-02-07
feat(blob): add rate limited error

luismeyer:


packages/blob/src/api.ts
```diff
@@ -162,13 +173,41 @@ export async function requestApi<TResponse>(
...
+        if (retryAfterSeconds > 2 * 60) {
+          bail(retryError);
+          return;
+        }
+
+        await new Promise<void>((resolve) => {
+          debug(
+            `waiting for ${retryAfterSeconds} seconds because of rate limit`,
+          );
+          setTimeout(resolve, 1000 * retryAfterSeconds);
```
correttojs: note that the `retryAfterSeconds` will be added to the exponential backoff introduced by `async-retry`

packages/blob/src/api.ts
```diff
@@ -162,13 +173,41 @@ export async function requestApi<TResponse>(
...
+      // only retry for certain errors
       if (code === 'unknown_error' || code === 'service_unavailable') {
-        // only retry for certain errors
         throw error;
-      } else {
-        // don't retry for e.g. suspended stores
-        bail(error);
       }
+
+      if (code === 'rate_limited') {
```
correttojs: I wonder if it's worth adding all this logic instead of playing with the https://www.npmjs.com/package/async-retry options (i.e. increase the `minTimeout` or `randomize` or `factor`). 
It would be nice if we could update these options at runtime

luismeyer: it seems this is not possible with `aync-retry` specifically because of the `node-retry` implementation. They calculate the timeouts once before they retries are startet. Therefor it's not possible to change during runtime. I'm wondering if it might be worth it to implement the retry mechanism our selfs, then we won't be limited by `node-retry`

packages/blob/src/api.ts
luismeyer: true...

packages/blob/src/api.ts
luismeyer: https://github.com/tim-kos/node-retry/blob/11efd6e4e896e06b7873df4f6e187c1e6dd2cf1b/lib/retry.js#L4

correttojs: I wouldn't implement our own retry, but rather simplify this logic.
Actually I see in `api` we have a similar implementation
```
 if (response.status === 429) {
            const retryAfter = parseInt(
              response.headers.get('retry-after') || '',
              10,
            );

            if (retryAfter && !isNaN(retryAfter)) {
              if (retryAfter > retryOpts.maxRetryAfter) {
                return response;
              }

              await sleep(retryAfter * 1e3);
              throw new ResponseError(response);
            }
          }
```
then I think what you did is good!
Not sure if you want to handle the `maxRetryAfter` and `isNaN`

correttojs (approved):

packages/blob/src/api.ts
luismeyer: I removed the automatic retries for rate limited errors completely for now. I think this is better than creating too long timeouts

correttojs (approved):

PR #612, opened 2024-02-18
make tracing optional

dferber90:


dferber90:
This approach will not work as edge runtime does not support top-level await.

PR #620, opened 2024-02-22
[edge-config] Make `@opentelemetry/api` optional and add `setTracerProvider` function

### LINKED ISSUE #611: `@vercel/edge-config` still breaks even if `@opentelemetry/api` is marked as peer deps
Continuing from #606 and #607, `@vercel/edge-config` still breaks even if `@opentelemetry/api` is marked as peer deps: https://vercel.com/grikomsn/roe/HZvq2b5jhhSTWaXNwGaBKUeDYbQS#L23-L24

CleanShot 2024-02-18 at 05 30 19@2x

One "fix" is to stub and override `@opentelemetry/api` with an empty object/function, which I tested and works: https://github.com/grikomsn/roe/commit/da7a36cde2d284c56e2bcb2f1435e875ccedb40f

```js
exports.trace = {
  getTracer() {
    return null;
  },
};
```

As a general fix, I don't know the best way moving forward. Opening this issue for visibility and point of discussion.

AndyBitz:
Fixes #611

By removing the `trace` import, we make the `@opentelemetry/api` fully optional, avoiding errors at build time, when the package does not exist.

Since we have no way of importing the package dynamically, we have to expose the `setTracerProvider` function that consumers of the library can use to set the `trace` themselves whenever they init it.

The `@opentelemetry/api` will continue to be a `peerDependency` that's marked as optional since we make use if its types for the `setTracer` function.

I made this a `minor` as opposed to a `major`, considering we change the API, and it fixes tracing, as it's currently broken, depending on how you see it.

### DIFF
--- a/.changeset/popular-trees-buy.md
+++ b/.changeset/popular-trees-buy.md
@@ -1,0 +1,5 @@
+---
+"@vercel/edge-config": minor
+---
+
+Make `@opentelemetry/api` optional and expose a `setTracer` function
--- a/.changeset/popular-trees-buy.md
+++ b/.changeset/popular-trees-buy.md
@@ -2,4 +2,4 @@
 "@vercel/edge-config": minor
 ---
 
-Make `@opentelemetry/api` optional and expose a `setTracer` function
+Make `@opentelemetry/api` optional and expose a `setTracerProvider` function
--- a/packages/edge-config/README.md
+++ b/packages/edge-config/README.md
@@ -121,7 +121,7 @@ export const config = { runtime: 'edge' };
 
 ## OpenTelemetry Tracing
 
-The `@vercel/edge-config` package makes use of the OpenTelemetry standard to trace certain functions for observability. In order to enable it, use the function `setTracer` to set the `TraceAPI` that should be used by the SDK.
+The `@vercel/edge-config` package makes use of the OpenTelemetry standard to trace certain functions for observability. In order to enable it, use the function `setTracerProvider` to set the `TracerProvider` that should be used by the SDK.
 
 ```js
 import { setTracerProvider } from '@vercel/edge-config';
--- a/packages/edge-config/README.md
+++ b/packages/edge-config/README.md
@@ -119,6 +119,19 @@ export default (req) => {
 export const config = { runtime: 'edge' };
 ```
 
+## OpenTelemetry Tracing
+
+The `@vercel/edge-config` package makes use of the OpenTelemetry standard to trace certain functions for observability. In order to enable it, use the function `setTracer` to set the `TraceAPI` that should be used by the SDK.
+
+```js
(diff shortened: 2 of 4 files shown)


packages/edge-config/src/utils/tracing.ts
```diff
@@ -1,12 +1,23 @@
...
 
+// Use a symbol to avoid having global variable that is scoped to this file,
+// as it can lead to issues with cjs and mjs being used at the same time.
+const edgeConfigTraceSymbol = Symbol.for('@vercel/edge-config:global-trace');
+
+/**
+ * Allows setting the `@opentelemetry/api` tracer to generate traces
+ * for Edge Config related operations.
+ */
+export function setTracer(tracer: TraceAPI): void {
```
dferber90: Seems like the recommended pattern is to use `TracerProvider`?

https://opentelemetry.io/docs/concepts/instrumentation/libraries/#getting-a-tracer

dferber90 (changes requested):

grikomsn (approved):

.changeset/popular-trees-buy.md
```diff
@@ -0,0 +1,5 @@
+---
+"@vercel/edge-config": minor
+---
+
+Make `@opentelemetry/api` optional and expose a `setTracer` function
```
dferber90:
```suggestion
Make `@opentelemetry/api` optional and expose a `setTracerProvider` function
```

packages/edge-config/README.md
```diff
@@ -119,6 +119,19 @@ export default (req) => {
 export const config = { runtime: 'edge' };
 ```
 
+## OpenTelemetry Tracing
+
+The `@vercel/edge-config` package makes use of the OpenTelemetry standard to trace certain functions for observability. In order to enable it, use the function `setTracer` to set the `TraceAPI` that should be used by the SDK.
```
dferber90:
```suggestion
The `@vercel/edge-config` package makes use of the OpenTelemetry standard to trace certain functions for observability. In order to enable it, use the function `setTracerProvider` to set the `TracerProvider` that should be used by the SDK.
```

packages/edge-config/README.md
```diff
@@ -119,6 +119,19 @@ export default (req) => {
...
+
+```js
+import { setTracerProvider } from '@vercel/edge-config';
+import { trace } from '@opentelemetry/api';
+
+setTracerProvider(trace);
```
dferber90, line 130: Is it correct to pass `trace` when a TracerProvider is expected?

dferber90 (changes requested):

packages/edge-config/README.md
AndyBitz: Yes

dferber90 (approved):

PR #632, opened 2024-03-11
chore: allow supabase connection pooler uri.

thorwebdev:
Currently folks have to use modify their supabase connection string to make it work with drizzle in vercel edge functions, e.g. see https://github.com/supabase-community/create-t3-turbo/blob/main/.env.example#L8

By changing this check, the supabase connection pooler URI passes by default.


vvo:

PR #643, opened 2024-04-01
Use a keep-alive agent for @vercel/kv by default

cramforce:
See https://upstash.com/docs/oss/sdks/ts/redis/advanced#keepalive

This has yielding substantial benefits while testing on v0

I asked why Upstash doesn't have this on-by-default
https://vercel.slack.com/archives/C02Q9BA1F8C/p1712002622991319 even though their blog says they do
https://upstash.com/blog/keepalive-in-serverless


cramforce:
Converted to draft because there might be issues still compiling this for Edge

PR #647, opened 2024-04-03
fix(blob): align put types with API response

### LINKED ISSUE #633: @vercel/blob `PutBlobResult.contentType` can be undefined
The type PutBlobResult states that the `contentType` is always a string, but is might actually be undefined.

The type should reflect reality

luismeyer:
fixes https://github.com/vercel/storage/issues/633

### DIFF
--- a/.changeset/smart-steaks-suffer.md
+++ b/.changeset/smart-steaks-suffer.md
@@ -1,0 +1,5 @@
+---
+"@vercel/blob": patch
+---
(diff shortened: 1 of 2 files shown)


dferber90 (approved):

vvo (approved):
I wonder if we should, later on, enforce a content-type by default. Browsers will choose one no matter what.

PR #650, opened 2024-04-05
feat(blob): add abortSignal option to all methods

### LINKED ISSUE #639: @vercel/blob - is it possible to cancel uploading a file?
Currently - `put` method seems to be returning a promise when the upload is done.

In my case - it is possible that the user will cancel the operation in the middle. Is it possible to cancel file upload then, so there are no unwanted artifacts and uploads left and there is no wasted network traffic?

luismeyer:
Fixes https://github.com/vercel/storage/issues/639

### DIFF
--- a/.changeset/cold-seas-hunt.md
+++ b/.changeset/cold-seas-hunt.md
@@ -3,4 +3,18 @@
 "vercel-storage-integration-test-suite": patch
 ---
 
+# Add abortSignal
+
 Adds `abortSignal` option to all methods. This allows users to cancel requests using an AbortController and passing its signal to the operation.
+
+Here's how to use it:
+
+```ts
+// don't await here
+const promise = vercelBlob.put('canceled.txt', 'test', {
+  access: 'public',
+  abortSignal: abortController.signal,
+});
+
+abortController.abort();
+```
--- a/.changeset/cold-seas-hunt.md
+++ b/.changeset/cold-seas-hunt.md
@@ -10,11 +10,17 @@ Adds `abortSignal` option to all methods. This allows users to cancel requests u
 Here's how to use it:
 
 ```ts
-// don't await here
-const promise = vercelBlob.put('canceled.txt', 'test', {
+const abortController = new AbortController();
+
+vercelBlob.put('canceled.txt', 'test', {
   access: 'public',
   abortSignal: abortController.signal,
+}).then(blob => {
+  console.log('Blob created:', blob);
 });
 
-abortController.abort();
+setTimeout(function() {
+  // Abort the upload
+  abortController.abort();
+}, 100);
 ```
--- a/.changeset/cold-seas-hunt.md
+++ b/.changeset/cold-seas-hunt.md
@@ -3,4 +3,4 @@
 "vercel-storage-integration-test-suite": patch
 ---
 
-Adds abortSignal option to all methods. This allows users to cancel requests by using an AbortController and passing it's signal to the operation.
+Adds `abortSignal` option to all methods. This allows users to cancel requests using an AbortController and passing its signal to the operation.
--- a/.changeset/cold-seas-hunt.md
+++ b/.changeset/cold-seas-hunt.md
@@ -1,0 +1,6 @@
+---
+"@vercel/blob": minor
+"vercel-storage-integration-test-suite": patch
+---
+
+Adds abortSignal option to all methods. This allows users to cancel requests by using an AbortController and passing it's signal to the operation.
--- a/packages/blob/src/api.ts
+++ b/packages/blob/src/api.ts
@@ -55,6 +55,12 @@ export class BlobServiceRateLimited extends BlobError {
   }
 }
 
+export class BlobRequestAbortedError extends BlobError {
+  constructor() {
+    super('The request was aborted.');
+  }
+}
+
 type BlobApiErrorCodes =
   | 'store_suspended'
   | 'forbidden'
@@ -196,7 +202,8 @@ export async function requestApi<TResponse>(
         });
       } catch (error) {
         if (error instanceof DOMException && error.name === 'AbortError') {
-          bail(error);
+          bail(new BlobRequestAbortedError());
+          return;
         }
 
         throw error;
--- a/packages/blob/src/api.ts
+++ b/packages/blob/src/api.ts
@@ -183,15 +183,24 @@ export async function requestApi<TResponse>(
 
   const apiResponse = await retry(
     async (bail) => {
-      const res = await fetch(getApiUrl(pathname), {
-        ...init,
-        headers: {
-          'x-api-version': apiVersion,
-          authorization: `Bearer ${token}`,
+      let res: Response;
+      try {
+        res = await fetch(getApiUrl(pathname), {
+          ...init,
+          headers: {
+            'x-api-version': apiVersion,
+            authorization: `Bearer ${token}`,
 
-          ...init.headers,
-        },
-      });
+            ...init.headers,
+          },
+        });
+      } catch (error) {
+        if (error instanceof DOMException && error.name === 'AbortError') {
+          bail(error);
+        }
+
+        throw error;
+      }
 
       if (res.ok) {
         return res;
--- a/packages/blob/src/api.ts
+++ b/packages/blob/src/api.ts
@@ -190,6 +190,8 @@ export async function requestApi<TResponse>(
   const apiResponse = await retry(
     async (bail) => {
       let res: Response;
+
+      // try/catch here to treat certain errors as not-retryable
       try {
         res = await fetch(getApiUrl(pathname), {
           ...init,
@@ -201,11 +203,13 @@ export async function requestApi<TResponse>(
           },
         });
       } catch (error) {
+        // if the request was aborted, don't retry
         if (error instanceof DOMException && error.name === 'AbortError') {
           bail(new BlobRequestAbortedError());
           return;
         }
 
+        // retry for any other erros thrown by fetch
         throw error;
       }
 
--- a/packages/blob/src/client.browser.test.ts
+++ b/packages/blob/src/client.browser.test.ts
@@ -133,12 +133,9 @@ describe('client', () => {
       const pathname = 'foo.txt';
       const token = 'vercel_blob_client_fake_token_for_test';
 
-      const abortController = new AbortController();
-
       const { uploadId, key } = await createMultipartUpload(pathname, {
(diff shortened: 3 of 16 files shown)


luismeyer:
fixes https://github.com/vercel/storage/issues/639

.changeset/cold-seas-hunt.md
```diff
@@ -0,0 +1,6 @@
+---
+"@vercel/blob": minor
+"vercel-storage-integration-test-suite": patch
+---
+
+Adds abortSignal option to all methods. This allows users to cancel requests by using an AbortController and passing it's signal to the operation.
```
vvo:
```suggestion
Adds `abortSignal` option to all methods. This allows users to cancel requests using an AbortController and passing its signal to the operation.
```

.changeset/cold-seas-hunt.md
```diff
@@ -0,0 +1,6 @@
+---
+"@vercel/blob": minor
+"vercel-storage-integration-test-suite": patch
+---
+
+Adds `abortSignal` option to all methods. This allows users to cancel requests using an AbortController and passing its signal to the operation.
```
vvo, line 8: @luismeyer Can we also add usage in the release notes? It's always pleasant to have everything in the changelog directly. Something like:

```
Here's how to use it (`put` example):

```

packages/blob/src/api.ts
```diff
@@ -183,15 +189,25 @@ export async function requestApi<TResponse>(
...
-        ...init,
-        headers: {
-          'x-api-version': apiVersion,
-          authorization: `Bearer ${token}`,
-
-          ...init.headers,
-        },
-      });
+      let res: Response;
+      try {
```
vvo, line 195: Just to be sure, can you comment here or in code why we now need a try catch? Pretty sure there's a good reason, just want to learn from it.

packages/blob/src/client.ts
```diff
@@ -27,6 +27,10 @@ export interface ClientCommonCreateBlobOptions {
    * Defines the content type of the blob. By default, this value is inferred from the pathname. Sent as the 'content-type' header when downloading a blob.
    */
   contentType?: string;
+  /**
+   * Abortsignal to cancel the running request.
```
```suggestion
   * `AbortSignal` to cancel the running request. See https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal
```

packages/blob/src/index.node.test.ts
```diff
@@ -566,5 +566,27 @@ describe('blob client', () => {
       });
       expect(headers['x-cache-control-max-age']).toEqual('60');
     });
+
+    const table: [string, (signal: AbortSignal) => Promise<unknown>][] = [
+      ['put', (s) => put('p', 'b', { access: 'public', abortSignal: s })],
```
vvo:
```suggestion
      ['put', (s) => put('file.txt', 'Hello World!', { access: 'public', abortSignal: s })],
```

packages/blob/src/index.node.test.ts
```diff
@@ -566,5 +566,27 @@ describe('blob client', () => {
...
+      ['del', (s) => del('path', { abortSignal: s })],
+      ['list', (s) => list({ abortSignal: s })],
+      ['copy', (s) => copy('f', 't', { access: 'public', abortSignal: s })],
```
vvo:
```suggestion
      ['copy', (s) => copy('file.txt', 'destination.txt', { access: 'public', abortSignal: s })],
```

packages/blob/src/index.node.test.ts
```diff
@@ -566,5 +566,27 @@ describe('blob client', () => {
       });
       expect(headers['x-cache-control-max-age']).toEqual('60');
     });
+
+    const table: [string, (signal: AbortSignal) => Promise<unknown>][] = [
+      ['put', (s) => put('p', 'b', { access: 'public', abortSignal: s })],
+      ['del', (s) => del('path', { abortSignal: s })],
```
vvo:
```suggestion
      ['del', (s) => del('file.txt', { abortSignal: s })],
```

packages/blob/src/index.node.test.ts
```diff
@@ -566,5 +566,27 @@ describe('blob client', () => {
...
+      ['list', (s) => list({ abortSignal: s })],
+      ['copy', (s) => copy('f', 't', { access: 'public', abortSignal: s })],
+      ['head', (s) => head('u', { abortSignal: s })],
```
vvo:
```suggestion
      ['head', (s) => head('file.txt', { abortSignal: s })],
```

packages/blob/src/multipart/complete.ts
```diff
@@ -72,6 +72,7 @@ export async function completeMultipartUpload({
           'x-mpu-key': encodeURI(key),
         },
         body: JSON.stringify(parts),
+        signal: options.abortSignal,
```
vvo, line 75: Curious: Why is this `options.abortSignal` and in list this is `options?.abortSignal`?

packages/blob/src/multipart/create-uploader.ts
```diff
@@ -25,6 +25,16 @@ export function createCreateMultipartUploaderMethod<
       options,
     );
 
+    // all sdk functions expose an abort signal but upload part requires a controller
+    let internalAbortController: AbortController | undefined;
+    if (options.abortSignal) {
+      internalAbortController = new AbortController();
+
+      options.abortSignal.addEventListener('abort', () => {
```
vvo: do we need to remove the listener when the request finished?

packages/blob/src/multipart/upload.ts
```diff
@@ -129,6 +129,11 @@ export function uploadAllParts({
   debug('mpu: upload init', 'key:', key);
   const internalAbortController = new AbortController();
 
+  // we connect the internal abort controller to the external abortSignal to allow the user to cancel the upload
+  options.abortSignal?.addEventListener('abort', () => {
+    internalAbortController.abort();
```
vvo: smart 👍

test/next/src/app/vercel/blob/script.mts
```diff
@@ -62,7 +67,9 @@ async function run(): Promise<void> {
 
   console.log(count, 'blobs in this store');
 
-  await Promise.all(urls.map((url) => vercelBlob.del(url)));
+  console.log(filteredUrls);
```
vvo: do we need your additions or just for debugging? I am fine to keep them

packages/blob/src/index.node.test.ts
vvo: why: always easier to know what we're dealing with (pathname or content), doesn't hurt too

packages/blob/src/multipart/upload.ts
vvo: Wondering @luismeyer shouldn't we instead do:
```
options.abortSignal?.addEventListener('abort', () => {
  cancel(); // abort correctly
});
```

?

packages/blob/src/multipart/complete.ts
luismeyer: all put operations require an options object to set `access: "public"`

packages/blob/src/multipart/upload.ts
luismeyer: yes makes sense

vvo: even options.abortSignal?.addEventListener('abort', cancel); should do it 👍

.changeset/cold-seas-hunt.md
```diff
@@ -0,0 +1,26 @@
...
+---
+
+# Add abortSignal
+
+Adds `abortSignal` option to all methods. This allows users to cancel requests using an AbortController and passing its signal to the operation.
+
+Here's how to use it:
+
+```ts
+const abortController = new AbortController();
```
vvo, line 13: @luismeyer I updated the example a bit further

packages/blob/src/client.browser.test.ts
```diff
@@ -133,9 +133,12 @@ describe('client', () => {
       const pathname = 'foo.txt';
       const token = 'vercel_blob_client_fake_token_for_test';
 
+      const abortController = new AbortController();
+
       const { uploadId, key } = await createMultipartUpload(pathname, {
         access: 'public',
         token,
+        abortSignal: abortController.signal,
```
vvo: Do we need this for the tests? I am confused as it seems we added an abortSignal to all the tests.

packages/blob/src/multipart/upload.ts
```diff
@@ -52,13 +50,13 @@ export function createUploadPartMethod<
...
 
-export function uploadPart({
+export async function uploadPart({
   uploadId,
   key,
   pathname,
   headers,
   options,
-  abortController,
+  abortController = new AbortController(),
```
vvo: Should we name this internalAbortController so we can easily distinguish it from a user abortController when reading the code?

packages/blob/src/multipart/upload.ts
```diff
@@ -90,6 +88,24 @@ export function uploadPart({
     },
     options,
   );
+
+  function handleAbort(): void {
+    abortController.abort();
```
vvo: If I pass a custom abortController to a multipart upload (put(..,.., {multipart: true, abortSignal: ..}), how do we ensure we call cancel() on abort? I am curious as I don't see it being done explicitly so perhaps it's done in another codepath

luismeyer: the cancel method lives inside the `uploadAllParts` function. I moved this abortion logic into the `uploadPart` function so that we don't have to do this signal connection step on the call side (in `uploadAllParts` and in the manual methods). In the `uploadAllParts` function we `try` the `uploadPart` call and invoke `cancel` in the `catch` block here: https://github.com/vercel/storage/blob/3c7d949fff61b73ed0fba7bc82cffa0e25539d3c/packages/blob/src/multipart/upload.ts#L306

vvo: so aborting will throw in uploadPart so that cancel is called, fine by me 👍 perhaps add a small comment at the line you linked so we understand in which case we could endup here

packages/blob/src/client.browser.test.ts
luismeyer: i changed it to expect an abortController only if it is really necessary. This is the case for uploadPart because there we always create an internalAbort Controller

vvo: 👍 got it now!

vvo (approved):

PR #657, opened 2024-04-15
chore(dx): fix manual script uniqueness of urls

vvo:
Some URLs were not unique, which made the script fail when multiple scripts instance were run in parallel.

### DIFF
--- a/test/next/src/app/vercel/blob/script.mts
+++ b/test/next/src/app/vercel/blob/script.mts
@@ -44,13 +44,8 @@ async function run(): Promise<void> {
 
   await Promise.all(
     urls.map(async (url) => {
-      try {
-        const blobDetails = await vercelBlob.head(url);
-        console.log(blobDetails, url);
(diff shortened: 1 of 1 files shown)


test/next/src/app/vercel/blob/script.mts
```diff
@@ -39,10 +39,18 @@ async function run(): Promise<void> {
...
   await Promise.all(
     urls.map(async (url) => {
-      const blobDetails = await vercelBlob.head(url);
-      console.log(blobDetails, url);
+      try {
+        const blobDetails = await vercelBlob.head(url);
+        console.log(blobDetails, url);
+      } catch (err) {
+        console.log('url', err);
+        process.exit(1);
```
luismeyer: why do we exit here? I think the script should still try to delete the files.

vvo: I can remove that, used only for testing

PR #665, opened 2024-04-23
Handle promise rejection

wyattjoh:
This handles the promise rejection when the `latestRequest` rejects. It's important to add where we used to have the `void` as invoking the `.then` will wait another microtick as well as re-throw any error that occurred during the initial promise. The promise is returned later, and ensures that the promise can be await'ed and/or handled.

Note, you may need to turn off hide whitespace to see the changes, I swapped out the `??=` to make it more explicit as to when the initialization of the promises occurred.

### DIFF
--- a/packages/edge-config/src/index.ts
+++ b/packages/edge-config/src/index.ts
@@ -136,48 +136,55 @@ function createGetInMemoryEdgeConfig(
   let latestRequest: Promise<EmbeddedEdgeConfig | null> | null = null;
 
   return trace(
-    async () => {
-      if (!shouldUseDevelopmentCache) return null;
+    () => {
+      if (!shouldUseDevelopmentCache) return Promise.resolve(null);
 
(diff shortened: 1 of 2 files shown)


packages/edge-config/src/index.ts
```diff
@@ -136,48 +136,55 @@ function createGetInMemoryEdgeConfig(
...
+          }
+
+          return { digest, items: body } as EmbeddedEdgeConfig;
+        });
 
-      // Ensure we don't keep a rejected promise in memory
-      embeddedEdgeConfigPromise.catch(() => {
-        embeddedEdgeConfigPromise = null;
-      });
+        // Once the request is resolved, we se the proper config to the promise
```
dferber90:
```suggestion
        // Once the request is resolved, we set the proper config to the promise
```

dferber90 (approved):

PR #670, opened 2024-04-30
fix(blob): better typing for `put()`

charlypoly:
Ensure to raise a type error when a file is created (not a folder) and the options are missing.

I just got started with Vercel Blob and ended up having a "missing options, see usage" BlobError that required me to take a look at the typing to understand that a 3rd argument was missing ([as stated in the docs](https://vercel.com/docs/storage/vercel-blob/using-blob-sdk#put)).

This is an attempt to strengthen the typing, some type predicates could be used to better narrow down the `optionsInput` inside the function's body.


luismeyer:
thank you for the contribution. 
We plan on changing the `put` interface in the future to only support files again. We want to introduce another function that is solely for creating folders. So this change will be removed very soon again and therefor I wouldn't create a new release for this

PR #675, opened 2024-05-05
chore(blob): add support for React Native in package.json

ghost:
https://nodejs.org/docs/latest-v19.x/api/packages.html#community-conditions-definitions


vvo:
@trentcharlie thanks! Were you able to verify this is working good in a react native app on your side? Thanks!

PR #680, opened 2024-05-17
feat(kv): enable auto pipelining by default

vvo:
We're making this a major release for safety but we believe
most applications can upgrade from 1.0.1 to 2 without any changes.
Auto pipelining should work by default and improve performance.

BREAKING CHANGE: Auto pipelining is on by default now, see
https://upstash.com/docs/oss/sdks/ts/redis/pipelining/auto-pipeline. This
brings performance benefits to any code making multiple redis commands
simultaneously.

If you detect bugs because of this, please open them at
https://github.com/vercel/storage/issues.

You can disable this new behavior with:
```js
import { createClient } from '@vercel/kv';

const kv = createClient({
  url: ..,
  token: ..,
  enableAutoPipelining: false
});
```

### DIFF
--- a/.changeset/rare-poems-try.md
+++ b/.changeset/rare-poems-try.md
@@ -2,4 +2,25 @@
 "@vercel/kv": major
 ---
 
-Enable auto pipelining by default
+We're making this a major release for safety but we believe
+most applications can upgrade from 1.0.1 to 2 without any changes.
+Auto pipelining should work by default and improve performance.
+
+BREAKING CHANGE: Auto pipelining is on by default now, see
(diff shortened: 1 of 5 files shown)


luismeyer (approved):

CahidArda:
Great to see auto pipelining in vercel storage!

We have a patch for auto-pipelining JSON commands in upstash-redis, we will release version 1.31.3 soon.

CahidArda:
[upstash-redis 1.31.3 is released](https://www.npmjs.com/package/@upstash/redis)

vvo:
@CahidArda updated 👍

cramforce (approved):

PR #734, opened 2024-08-02
feat(blob-local): add local Blob API for development

luismeyer:


.github/workflows/release.yml
```diff
@@ -43,3 +45,50 @@ jobs:
...
+        with:
+          # This configures an alias for `docker build` to run `docker buildx`
+          # so we don't have to update our build script to be platform-aware.
+          install: true
+          platforms: linux/amd64,linux/arm64
+
+      - name: Login to Docker Hub
+        uses: docker/login-action@v2
+        with:
+          password: ${{ secrets.DOCKER_PASSWORD }}
```
luismeyer, line 80: these need to be added

.changeset/config.json
```diff
@@ -10,5 +10,9 @@
   "ignore": [],
   "snapshot": {
     "useCalculatedVersion": true
+  },
+  "privatePackages": {
```
luismeyer, line 14: All public packages get published automatically. We don't want that for blob-local because it's not published to npm. Instead blob-local is a private package that we can create changesets for. If a changeset is created we will create a release and a tag but not publish

vvo:

PR #735, opened 2024-08-13
[edge-config] Add `consistentRead` option to client

AndyBitz:
Fixes FLA-1256

Adds `consistentRead` to the `get`, `has`, `getAll`, and `digest` function which allows bypassing all optimizations to read from the origin with the downside of increased latency.


AAorris (approved):

packages/edge-config/src/index.ts
```diff
@@ -195,12 +198,22 @@ function createGetInMemoryEdgeConfig(
   );
 }
 
+/**
```
dferber90, line 202: empty jsdoc

packages/edge-config/src/index.ts
```diff
@@ -262,6 +275,15 @@ interface EdgeConfigClientOptions {
   cache?: 'no-store' | 'force-cache';
 }
 
+interface EdgeConfigFunctionsOptions {
+  /**
+   * Enabling `consistentRead` will bypass all caches and hit the origin
+   * directly. This will make sure to fetch the most recent version of
+   * an Edge Config with the downside of an increased latency.
```
dferber90:
```suggestion
   * an Edge Config with the downside of an increased latency.
   *
   * We do not recommend enabling this option, unless you are reading Edge Config specifically for generating a page using ISR and you need to ensure you generate with the latest content.
```

dferber90 (approved):

dferber90 (approved):

.changeset/polite-taxis-repair.md
```diff
@@ -0,0 +1,5 @@
+---
+'@vercel/edge-config': minor
+---
+
+Add the `consistentRead` option to allow reading from the origin
```
dferber90: Let's be clear this is not recommended even here in the changelog, as it will have way higher latency and can not handle the same read volume.

PR #744, opened 2024-09-06
feat(postgres): remove utf-8-validate dependency and update minimum node version

lukahartwig:
`utf-8-validate` is a polyfill that is no longer needed in Node 18.14


adriancooney (approved):
Happy to approve and acknowledge this is a breaking change. As we discussed, we are below 1.0 so breaking changes are expected in minor versions and in the unlikely case this breaks someone's build (since this is targeted at serverless where the minimum version is 18), we can always revert.

PR #746, opened 2024-09-06
feat(postgres): remove bufferutil dependency

lukahartwig:
`bufferutil` is an optional dependency to improve performance of the `ws` package (see [here](https://github.com/websockets/ws#opt-in-for-performance)). However, the performance benefit is likely not that big due to the network overhead and since it is a native package it can introduce friction in the build process.

Users of `@vercel/postgres` can still opt-in by installing `bufferutil` manually.

### DIFF
--- a/.changeset/slimy-walls-thank.md
+++ b/.changeset/slimy-walls-thank.md
@@ -1,0 +1,5 @@
+---
+'@vercel/postgres': minor
+---
+
+Removes optional bufferutil dependency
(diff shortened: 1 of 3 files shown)


vvo:
@lukahartwig do we still need this?

lukahartwig:
> @lukahartwig do we still need this?

According to the documentation `bufferutil` is supposed to be an optional dependency but removing it seems to break the tests. Haven't had the time to look into this yet.

I'll close the PR and re-open when I have the time to investigate.

PR #750, opened 2024-09-12
fix(blob): limit keys to 950 characters

vvo:
To respect internal limitations we have, let's limit keys to 950 charactersa and
warn when you use more than that.

### DIFF
--- a/packages/blob/src/put-helpers.ts
+++ b/packages/blob/src/put-helpers.ts
@@ -68,6 +68,12 @@ export function createPutHeaders<TOptions extends CommonPutCommandOptions>(
   return headers;
 }
 
+// maximum path length is:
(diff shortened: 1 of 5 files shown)


packages/blob/src/put-helpers.ts
```diff
@@ -85,6 +91,12 @@ export async function createPutOptions<
     throw new BlobError('pathname is required');
   }
 
+  if (pathname.length > MAXIMUM_PATHNAME_LENGTH) {
```
luismeyer, line 89: is this function also used by copy? If not we need to add the check to the copy method aswell

vvo: indeed, I'll add that now

vvo: Added in cdc40b753adb9f4610e37beb3d52291c517f0eab

luismeyer (approved):

Issue #640, opened 2024-03-26
@vercel/blob - Folder entry coming through as blob


Kvek:
Hi,
When fetching the blobs using the `list()`, I noticed the sub folder (`trending`) containing the data coming through as an entry.

Folder structure in dashboard

> Blob project 
>> trending
>>> ...images

This happens for both `folded` and the default `expanded` modes. 

config
```
import { list } from "@vercel/blob";

  const { blobs } = await list({
    mode: "folded",
    prefix: "trending/"
  });
```

![Screenshot 2024-03-26 at 20 44 41](https://github.com/vercel/storage/assets/8972543/cdb43630-af9c-4fe8-9476-bfc5f2abc653)

![Screenshot 2024-03-26 at 20 45 15](https://github.com/vercel/storage/assets/8972543/1b6b784c-f2af-4a53-bc70-2f49d44cfde6)

Is this intended behaviour?

<hr>
I also receive this error presumably because the blob's url entry isn't a valid image url.

```
The requested resource isn't a valid image for https://3gzqlfvwgaygv21a.public.blob.vercel-storage.com/trending/ 
received application/octet-stream
``` 

Thanks.

gustavo-bonfim:
Same here

PR #751, opened 2024-09-13
feat(blob): add createFolder method

vvo:
And remove variadic calls from main put methods, which will allow us to more
easily upgrade to newer TS versions.

Related:
- https://github.com/vercel/storage/pull/682
- https://github.com/vercel/storage/pull/667

### DIFF
--- a/packages/blob/src/create-folder.ts
+++ b/packages/blob/src/create-folder.ts
@@ -1,0 +1,44 @@
+import { requestApi } from './api';
+import type { BlobCommandOptions } from './helpers';
+import { putOptionHeaderMap, type PutBlobApiResponse } from './put-helpers';
+
+export interface CreateFolderResult {
+  pathname: string;
+  /**
+   * URL of created folder. There is not content at this URL.
+   * Use this URL to delete the folder, just like you would delete a file.
+   */
+  url: string;
+}
+
+/**
+ * Creates a folder in your store. Vercel Blob has no real concept of folders, our file browser on Vercel.com displays folders based on the presence of trailing slashes in the pathname. Unless you are building a file browser system, you probably don't need to use this method.
+ * @param pathname - Can be user1/ or user1/avatars/
+ * @param options - Additional options like `token`
+ * @returns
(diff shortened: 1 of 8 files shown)


packages/blob/src/create-folder.ts
```diff
@@ -0,0 +1,44 @@
...
+  pathname: string;
+  /**
+   * URL of created folder. There is not content at this URL.
+   * Use this URL to delete the folder, just like you would delete a file.
+   */
+  url: string;
+}
+
+/**
+ * Creates a folder in your store. Vercel Blob has no real concept of folders, our file browser on Vercel.com displays folders based on the presence of trailing slashes in the pathname. Unless you are building a file browser system, you probably don't need to use this method.
```
luismeyer, line 11: should we warn here, that folders will appear as blobs when you call `list()`? https://github.com/vercel/storage/issues/640

test/next/src/app/vercel/blob/script.mts
```diff
@@ -18,26 +18,26 @@ console.log();
...
-    axiosExample(),
-    gotExample(),
-    fetchExample(),
-    noExtensionExample(),
-    weirdCharactersExample(),
-    copyTextFile(),
-    listFolders(),
-    multipartNodeJsFileStream(),
-    fetchExampleMultipart(),
+    // textFileExample(),
```
luismeyer: this is for testing I guess

packages/blob/src/put.ts
```diff
@@ -54,14 +43,14 @@ export function createPutMethod<TOptions extends PutCommandOptions>({
     const options = await createPutOptions({
       pathname,
       // when no body is required (for folder creations) options are the second argument
```
luismeyer: this can be removed

luismeyer (approved):

test/next/src/app/vercel/blob/script.mts
vvo: yes will remove

PR #777, opened 2024-10-11
fix(blob): provide custom errors for expired client tokens and pathname mismatch

### LINKED ISSUE #456: Error: Vercel Blob: Access denied, please provide a valid token for this resource
Im trying to upload an image file using client upload method using pages route code given in [vercel documentation](https://vercel.com/docs/storage/vercel-blob/quickstart#client-uploads). Client token is being generated successfully but the upload fails with 403.

Im using:

"next": "^13.5.2", 
"@vercel/blob": "^0.14.1",

vvo:
Before this commit, when using client uploads and experiencing errors like expired token or pathname mismatch, it would be impossible to know the exact reason. This was frequently reported as time wasted by customers.

We're now exposing:
- BlobPathnameMismatchError
- BlobClientTokenExpiredError
- BlobFileTooLargeError

Fixes #456
Fixes STO-1710


vvo:
Adding one more custom error (file size)

luismeyer (approved):

PR #782, opened 2024-10-24
feat(blob): Provide onUploadProgress({ loaded, total, percentage })

vvo:
This commit introduces an `onUploadProgress` callback to put/upload*. Here's how to use it:

```ts
// also works with upload()
const blob = await put('file.pdf', file, {
  onUploadProgress(event) {
    console.log(event.loaded, event.total, event.percentage);
  }
});
```

**How does it work?**

By default we use `fetch`. When you pass `onUploadProgress`, we turn the file into a stream to track the stream consumption's progress by fetch.
For browsers not supporting `fetch` with streams (Safari, Firefox), we use `XMLHttpRequest` and its progress events.

This means we re-implemented parts of https://github.com/axios/axios or other fetch libraries using xhr as backup. But remember, progress events are somehow lying as they only measure what gets sent from your computer to the server, not what the server really acknowledged. Still, they are a good indicator for progress of the upload.


packages/blob/src/api.ts
```diff
@@ -252,10 +247,12 @@ async function getBlobError(
   return { code, error };
 }
 
+const CHUNK_SIZE = 64 * 1024;
+
 export async function requestApi<TResponse>(
   pathname: string,
   init: RequestInit,
-  commandOptions: BlobCommandOptions | undefined,
+  commandOptions: (BlobCommandOptions & WithUploadProgress) | undefined,
```
vvo, line 254: I initially thought onUploadProgress should be added to BlobCommandOptions, but BlobCommandOptions is shared between public and private interfaces for all blob commands. While onUploadProgress is available only for put/upload. If we add it to BlobCommandOptions then it will show as an optional options for all commands (even head)

packages/blob/src/api.ts
```diff
@@ -268,20 +265,104 @@ export async function requestApi<TResponse>(
...
+        // 1. For upload progress we always need to know the total size of the body
+        // 2. In development we need the header for put() to work correctly when passing a stream
+        (commandOptions?.onUploadProgress || shouldUseXContentLength())
+      ) {
+        bodyLength = computeBodyLength(init.body);
+      }
+
+      if (init.body) {
+        if (commandOptions?.onUploadProgress) {
+          if (supportsRequestStreams) {
```
luismeyer: why do we need this multilevel if instead of `&&`

packages/blob/src/api.ts
```diff
@@ -268,20 +265,104 @@ export async function requestApi<TResponse>(
...
+            // So that on retries we can reuse the original body, otherwise we would not be able to reuse it
+            const stream = await toReadableStream(init.body as PutBody);
+
+            let loaded = 0;
+
+            const chunkTransformStream = createChunkTransformStream(
+              CHUNK_SIZE,
+              (newLoaded: number) => {
+                loaded += newLoaded;
+                const total = bodyLength ?? loaded;
```
luismeyer: this optional chaining is very confusing to me. This is just to make TS happy right? Could we add bodyLength to the if check to make it clearer that we always know the bodyLength here

packages/blob/src/api.ts
```diff
@@ -268,20 +265,104 @@ export async function requestApi<TResponse>(
...
+                const percentage = Number(((loaded / total) * 100).toFixed(2));
+
+                // Leave percentage 100 to end of request
+                if (percentage === 100) {
+                  return;
+                }
+
+                commandOptions.onUploadProgress?.({
```
luismeyer, line 321: the optional chaining is unnecessary

packages/blob/src/api.ts
```diff
@@ -268,20 +265,104 @@ export async function requestApi<TResponse>(
...
+                  loaded,
+                  // When passing a stream to put(), we have no way to know the total size of the body.
```
luismeyer, line 323: > 1. For upload progress we always need to know the total size of the body

isn't this contradicting

packages/blob/src/helpers.ts
```diff
@@ -83,3 +107,108 @@ export function isPlainObject(value: unknown): boolean {
...
+// Microsoft Edge: implemented (Chromium)
+// Firefox: not implemented, BOO!! https://bugzilla.mozilla.org/show_bug.cgi?id=1469359
+// Safari: not implemented, BOO!! https://github.com/WebKit/standards-positions/issues/24
+export const supportsRequestStreams = (() => {
+  let duplexAccessed = false;
+
+  const hasContentType = new Request(getApiUrl(), {
+    body: new ReadableStream(),
+    method: 'POST',
+    // @ts-expect-error -- duplex option exists
```
luismeyer: is this due to an outdated TS version? Should we update?

packages/blob/src/helpers.ts
```diff
@@ -83,3 +107,108 @@ export function isPlainObject(value: unknown): boolean {
...
+    // noop
+  }
+  return `${baseUrl || 'https://blob.vercel-storage.com'}${pathname}`;
+}
+
+const TEXT_ENCODER =
+  typeof TextEncoder === 'function' ? new TextEncoder() : null;
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any -- We want to feature detect
+export function computeBodyLength(body: any): number {
```
luismeyer: can we use `object` to avoid `any`

packages/blob/src/helpers.ts
```diff
@@ -83,3 +107,108 @@ export function isPlainObject(value: unknown): boolean {
...
+  /* eslint-disable @typescript-eslint/no-unsafe-member-access -- We want to feature detect */
+
+  if (typeof body === 'string') {
+    if (TEXT_ENCODER) {
+      return TEXT_ENCODER.encode(body).byteLength;
+    }
+
+    // React Native doesn't have TextEncoder
+    return new Blob([body]).size;
+  } else if (typeof body.byteLength === 'number') {
```
luismeyer: no need for `else if` because all branches exit early

packages/blob/src/helpers.ts
```diff
@@ -83,3 +107,108 @@ export function isPlainObject(value: unknown): boolean {
...
+      const newBuffer = new Uint8Array(buffer.length + chunk.byteLength);
+      newBuffer.set(buffer);
+      newBuffer.set(new Uint8Array(chunk), buffer.length);
+      buffer = newBuffer;
+
+      // Output complete chunks
+      while (buffer.length >= chunkSize) {
+        const newChunk = buffer.slice(0, chunkSize);
+        controller.enqueue(newChunk);
+        onProgress && onProgress(newChunk.byteLength);
```
luismeyer:
```suggestion
        onProgress?.(newChunk.byteLength);
```

packages/blob/src/helpers.ts
```diff
@@ -83,3 +107,108 @@ export function isPlainObject(value: unknown): boolean {
...
+        buffer = buffer.slice(chunkSize);
+      }
+    },
+
+    flush(controller) {
+      // Send any remaining data
+      if (buffer.length > 0) {
+        controller.enqueue(buffer);
+        onProgress && onProgress(buffer.byteLength);
```
luismeyer:
```suggestion
        onProgress?.(newChunk.byteLength);
```

packages/blob/src/multipart/upload.ts
```diff
@@ -174,6 +173,27 @@ export function uploadAllParts({
     let arrayBuffers: ArrayBuffer[] = [];
     let currentPartBytesRead = 0;
 
+    let onUploadProgress: (() => void) | undefined;
+    const totalLoadedPerPartNumber: Record<string, number> = {};
+
+    if (options.onUploadProgress) {
+      onUploadProgress = throttle(() => {
```
luismeyer, line 179: why do we throttle this? Doesn't this make the upload feel slower

luismeyer (changes requested):

packages/blob/src/api.ts
vvo: We cannot add it to the if I believe, since bodyLength can be `0` at this point (if you pass a stream as body). But we do not need ?? either, would using `bodyLength || loaded` be ok per your review?

packages/blob/src/api.ts
vvo: We do not 👍 fixed in 9a4a7fd45ccf2834ee09d28674639dde7eb7d727

packages/blob/src/api.ts
vvo: It seems required, probably because the call is too far away for TS to be sure it will still be here even after the if? 
CleanShot 2024-10-24 at 14 16 13@2x

I've added a local variable to keep a reference to it in 59e01f6192cff9f9e74ed61ff2ba8db63e427260

packages/blob/src/helpers.ts
vvo: Seems like no version of TS ships this at the moment: https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/1729

I added this link in the code now.

packages/blob/src/helpers.ts
vvo: I am unsure if this is doable when trying to use feature detection: 
CleanShot 2024-10-24 at 14 38 17@2x

Do you have any other idea (I'd rather avoid gigantic guards to make ts happy).

(this code is heavily inspired by https://github.com/smithy-lang/smithy-typescript/blob/main/packages/util-body-length-browser/src/calculateBodyLength.ts)

vvo: Ok here's the correct way 🙏 https://github.com/vercel/storage/pull/782/commits/6e6d7b90e9599271fab3b3c5de20bb3bd43b2a2b

packages/blob/src/helpers.ts
vvo: Good, updated, I wish this was linted too

packages/blob/src/multipart/upload.ts
vvo: I chose to throttle at 100ms as it looked like most libraries providing uploadProgress were also throttling sending events (example: https://github.com/axios/axios/blob/v1.x/lib/helpers/progressEventReducer.js).

I guess it's here to provide a better experience in most cases because otherwise, you can receive a LOT of progress events.

With throttling:

https://github.com/user-attachments/assets/49435109-67ec-4158-81ed-bca20520d40a

Without throttling:

https://github.com/user-attachments/assets/9ef3c97a-9146-45b6-949d-ebbcf42131d2

I am happy to change my mind around this if we think this could be bad for customers, it's also something we can just remove without breaking the API if needed (rare case!)

luismeyer: I see. First video looks way nicer 👍

packages/blob/src/helpers.ts
luismeyer: nice

packages/blob/src/api.ts
luismeyer: Ah I see. It's probably because it's within a callback

packages/blob/src/api.ts
vvo: moved to ||

packages/blob/src/helpers.ts
vvo: (applied)

packages/blob/src/helpers.ts
vvo: applied

luismeyer (approved):

PR #831, opened 2025-02-21
[@vercel/kv, @vercel/postgres, @vercel/postgres-kysely] Add notices to READMEs

adriancooney:
This PR adds notices to the READMEs of `@vercel/kv`, `@vercel/postgres` and `@vercel/postgres-kysely`. The Vercel KV and Vercel Postgres have transitioned over to Neon and Upstash in the Vercel Marketplace, unlocking the full potential of both platforms for Vercel customers. As a consequence, we are recommending `@vercel/kv` in favour of Upstash's `@upstash/redis` module and `@vercel/postgres` in favour of Neon's `@neondatabase/serverless`. These modules were essentially just wrappers around these modules anyway (with some added bells and whistles and API sprinkles). `npm` deprecation notices will follow soon.

### DIFF
--- a/packages/kv/README.md
+++ b/packages/kv/README.md
@@ -2,7 +2,7 @@
 
 <!-- prettier-ignore -->
 > [!CAUTION]
-> **`@vercel/kv` has moved to [the `@upstash/redis` package from Upstash](https://github.com/upstash/redis-js).
+> **`@vercel/kv` has moved to [the `@upstash/redis` package from Upstash](https://github.com/upstash/redis-js).** This retains the same first-party billing and integration into Vercel, with the added ability to open your Redis stores directly in Upstash.
 
 A client that works with Vercel KV.
 
--- a/packages/kv/README.md
+++ b/packages/kv/README.md
@@ -1,5 +1,9 @@
 # @vercel/kv
 
+<!-- prettier-ignore -->
+> [!CAUTION]
+> **`@vercel/kv` has moved to [the `@upstash/redis` package from Upstash](https://github.com/upstash/redis-js).
+
 A client that works with Vercel KV.
 
 ## Install
--- a/packages/postgres/README.md
+++ b/packages/postgres/README.md
@@ -1,5 +1,9 @@
 # @vercel/postgres
 
+<!-- prettier-ignore -->
+> [!CAUTION]
+> **`@vercel/postgres` has moved to the [`@neondatabase/serverless` package from Neon instead](https://github.com/neondatabase/serverless).** See the [Vercel Postgres to Neon transition guide](https://neon.tech/docs/guides/vercel-postgres-transition-guide).
+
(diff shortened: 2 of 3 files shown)


packages/kv/README.md
```diff
@@ -1,4 +1,6 @@
-# @vercel/kv
+# @vercel/kv (deprecated)
+
+> [!CAUTION] > **`@vercel/kv` is deprecated**. Please use [the `@upstash/redis` package from Upstash instead](https://github.com/upstash/redis-js).
```
correttojs: Screenshot 2025-02-21 at 12 05 16

it seems the Note didn't render correctly


Should we add some info about the features in this package not avilable in `@upstash/redis`?

packages/postgres/README.md
```diff
@@ -1,4 +1,6 @@
-# @vercel/postgres
+# @vercel/postgres (deprecated)
+
+> [!CAUTION] > **`@vercel/postgres` is deprecated**. Please use [the `@neondatabase/serverless` package from Neon instead](https://github.com/neondatabase/serverless). See the [Vercel Postgres to Neon transition guide](https://neon.tech/docs/guides/vercel-postgres-transition-guide).
```
correttojs: do we need to deprecate https://github.com/vercel/storage/tree/main/packages/postgres-kysely as well?

packages/kv/README.md
adriancooney: Grrr. My editor formatter removed the new line, very strange. Updated! Thanks for checking.

> Should we add some info about the features in this package not avilable in @upstash/redis?

What had you in mind?

correttojs: - this package overrides `scan`, `hscanIterator`, `sscan`, `zscan`...not sure if upstash fixed that issue in the meantime, otherwise we could mention it in the deprecation note

packages/postgres/README.md
adriancooney: yep, apologies! I think I pushed an older commit.

packages/kv/README.md
adriancooney: So you mean like an upgrade guide? I'll check in with the docs team to see where we are on that.

correttojs (approved):

packages/kv/README.md
```diff
@@ -1,5 +1,9 @@
 # @vercel/kv
 
+<!-- prettier-ignore -->
+> [!CAUTION]
+> **`@vercel/kv` has moved to [the `@upstash/redis` package from Upstash](https://github.com/upstash/redis-js).
```
adriancooney:
```suggestion
> **`@vercel/kv` has moved to [the `@upstash/redis` package from Upstash](https://github.com/upstash/redis-js).** This retains the same first-party billing and integration into Vercel, with the added ability to open your Redis stores directly in Upstash.
```

PR #837, opened 2025-02-28
chore(deps): upgrade deps

vvo:


packages/postgres/package.json
```diff
@@ -49,19 +49,19 @@
     "ws": "^8.17.1"
   },
   "devDependencies": {
-    "@changesets/cli": "2.27.11",
```
correttojs: can we skip postgres, kysley, kv as they are deprecated?
it would be confusing for the user to see a new release

vvo: There won't be a new release on them, this was only dev deps, I refrained myself from upgrading any deps on kv/postgres that would require a new release. Is that ok or do you want me to remove that? Let me know

correttojs: if possible I'd avoid any update to these packages...the user may still check the code repo and see some discrepancies. It may also be useful for us to know at which dependency version we deprecated the package

vvo: 👍 gonna remove it

vvo: I removed the changes for kv/postgres* @correttojs now 👍

correttojs (approved):

PR #838, opened 2025-03-13
fix(blob): Correct README and types links to Blob documentation

vvo:
No functional changes to the package. Only documentation and types update to correct links to Vercel doc.

### DIFF
--- a/packages/blob/CHANGELOG.md
+++ b/packages/blob/CHANGELOG.md
@@ -421,7 +421,7 @@
 
 - 8251462: This release introduces BREAKING CHANGES. Mostly, we've separated client and server needs better. While ensuring we only export what we currently think is useful to you.
 
-  We have a completely new documentation about client (browser) uploads: https://vercel.com/docs/storage/vercel-blob/quickstart#client-uploads.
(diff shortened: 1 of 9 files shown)


packages/blob/CHANGELOG.md
```diff
@@ -421,7 +421,7 @@
 
 - 8251462: This release introduces BREAKING CHANGES. Mostly, we've separated client and server needs better. While ensuring we only export what we currently think is useful to you.
 
-  We have a completely new documentation about client (browser) uploads: https://vercel.com/docs/storage/vercel-blob/quickstart#client-uploads.
+  We have a completely new documentation about client (browser) uploads: https://vercel.com/docs/vercel-blob/quickstart#client-uploads.
```
molebox: This is incorrect still. We dont have a single quickstart, we have one for client and one for server uploads

molebox (changes requested):

packages/blob/CHANGELOG.md
vvo: right!

vvo: Done @molebox

PR #854, opened 2025-05-19
fix(blob): Use new endpoint for API calls

### LINKED ISSUE #853: Failed to load resource: net::ERR_FAILED
i got a lot `Failed to load resource: net::ERR_FAILED` error after updating to `@vercel/blob@1.0.1`

vvo:
### TL;DR

Updated Vercel Blob API endpoint URL and version to use the new API path structure.

### What changed?

- Changed the Blob API base URL from `https://blob.vercel-storage.com` to `https://vercel.com/api/blob`
- Updated the API version from `10` to `11`
- Updated all API endpoint paths in tests to reflect the new URL structure
- Modified test assertions to match the new endpoint patterns

### Why make this change?

This change aligns the Blob API with Vercel's standard API URL structure, consolidating all Vercel APIs under the same domain. The version bump to `11` ensures compatibility with the new endpoint structure and allows the server to properly handle requests from updated clients.

We've also seen cases where requests were failing and using vercel.com/api solves that.

fixes #853

### DIFF
--- a/packages/blob/src/api.node.test.ts
+++ b/packages/blob/src/api.node.test.ts
@@ -35,7 +35,7 @@ describe('api', () => {
       process.env.BLOB_READ_WRITE_TOKEN = undefined;
 
       await expect(
-        requestApi('/api', { method: 'GET' }, undefined),
+        requestApi('/method', { method: 'GET' }, undefined),
       ).rejects.toThrow(BlobError);
 
       expect(fetchMock).toHaveBeenCalledTimes(0);
@@ -51,7 +51,7 @@ describe('api', () => {
       );
 
       const res = await requestApi<{ success: boolean }>(
-        '/api',
+        '/method',
         { method: 'POST', body: JSON.stringify({ foo: 'bar' }) },
         { token: '123' },
       );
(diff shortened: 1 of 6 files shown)


vvo:
* **#854** <a href="https://app.graphite.dev/github/pr/vercel/storage/854?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/vercel/storage/854?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

packages/blob/src/api.node.test.ts
```diff
@@ -51,22 +51,22 @@ describe('api', () => {
...
-        'https://blob.vercel-storage.com/api',
+        'https://vercel.com/api/blob/method',
         {
           body: '{"foo":"bar"}',
           headers: {
             authorization: 'Bearer 123',
             'x-api-blob-request-attempt': '0',
             'x-api-blob-request-id': expect.any(String) as string,
-            'x-api-version': '10',
+            'x-api-version': '11',
```
luismeyer, line 69: why is this an api version change?

vvo: I want to ensure we can track down this update in metrics (i.e. compare retries by version for example)

luismeyer (approved):

PR #878, opened 2025-09-26
feat(@vercel/edge-config) Enforce an IO boundary around client APIs

gnoff:
Edge config is an IO library but it heavily caches the underlying IO to make reads incredibly fast. Next.js has a heuristic for prerenders that excludes IO. While we could argue that the kind of data edge-config provides is prerenderable because it has a long lifetime it is not clear that users expect this. To keep edge-config reads consistently treated as IO I've added a wrapper for the API methods which guarantees that any read will resolve in a future timeout task.

The choice of timeout is to align with React's IO semantics for what is and is not considered IO for purposes of Suspending. On edge runtimes this is unthrottled and should not add meaningful overhead. In Node.js this is at a minimum a 1ms delay if no other work takes time however since the edge config reads themselves may take longer than 1 ms often it won't be observed. Additionally as soon as one read is queued all subsequent reads that are initiated before the timeout executes will share the same boundary.

There is of course the possibility that previously scheduled events like immediates and OS IO can delay the unblocking of these APIs but if the workload is already CPU bound then reordering the resolution is not likely to meaningfully affect overall performance.

### Best reviewed with [whitespace ignored](https://github.com/vercel/storage/pull/878/files?w=1)

### DIFF
--- a/packages/edge-config/src/utils/io-boundary.ts
+++ b/packages/edge-config/src/utils/io-boundary.ts
@@ -1,0 +1,50 @@
+let blockedBoundary: null | Promise<void> = null;
+
+function getCurrentBoundary(): Promise<void> {
+  if (blockedBoundary) {
+    return blockedBoundary;
+  }
+  blockedBoundary = new Promise((resolve) => {
+    // We use setTimeout intentionally even in Node.js environments to ensure these
+    // are tracked as IO by React in Server Component environments.
+    // TODO implement a Next.js specific version so we can skip this boundary checking
+    // except where needed.
+    setTimeout(() => {
+      blockedBoundary = null;
+      resolve();
+    }, 0);
+  });
+  return blockedBoundary;
+}
+
+/**
+ * Wraps an async function to ensure that it will always yield to the event loop before
+ * resolving or rejecting.
+ */
+export function withIOBoundary<
+  /* eslint-disable @typescript-eslint/no-explicit-any -- we're wrapping */
+  F extends (...args: any[]) => Promise<any>,
+>(fn: F): F {
+  async function bounded(this: unknown, ...args: unknown[]): Promise<unknown> {
+    const boundary = getCurrentBoundary();
+    try {
+      /* eslint-disable @typescript-eslint/no-unsafe-return -- we're wrapping */
+      return await fn.call(this, ...args);
+    } finally {
+      await boundary;
+    }
+  }
+
+  try {
+    Object.defineProperty(bounded, 'name', { value: fn.name });
+  } catch {
+    // best effort
+  }
+  try {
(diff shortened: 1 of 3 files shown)


packages/edge-config/src/utils/io-boundary.ts
```diff
@@ -0,0 +1,50 @@
...
+function getCurrentBoundary(): Promise<void> {
+  if (blockedBoundary) {
+    return blockedBoundary;
+  }
+  blockedBoundary = new Promise((resolve) => {
+    // We use setTimeout intentionally even in Node.js environments to ensure these
+    // are tracked as IO by React in Server Component environments.
+    // TODO implement a Next.js specific version so we can skip this boundary checking
+    // except where needed.
+    setTimeout(() => {
```
dferber90, line 12: > however since the edge config reads themselves may take longer than 1 ms often it won't be observed

We worked very hard to make latency <1ms in the majority of Edge Config reads, as we don't go over the network. p75 is definitely under 1ms.

With this change we would **_always_** be over 1ms in observed latency if any user measures `await get("foo")`. That's quite the loss.

Is there any other way to signal an IO operation to Next.js that does not require artificially delaying the execution of the operation?

And if we absolutely must merge this I would prefer skipping the setTimeout for non-Next.js to avoid punishing other frameworks for what smells a bit like Next.js leaking its implementation details to this SDK.

Can we find another way of achieving this that does not rely on the heuristic?

packages/edge-config/src/utils/io-boundary.ts
```diff
@@ -0,0 +1,50 @@
+let blockedBoundary: null | Promise<void> = null;
+
```
dferber90, line 2: _This comment is in random place so we can have a threaded discussion_

> While we could argue that the kind of data edge-config provides is prerenderable because it has a long lifetime it is not clear that users expect this.

I'm fully on board that the default expectation is for this to be IO and not part of the prerender. For, [we have this explicit `no-store` configuration option](https://github.com/vercel/storage/blob/main/packages/edge-config/src/index.ts#L278), that is used by some users of this library to treat Edge Config more like a CMS. Would that still work with the proposed changes?

dferber90 (changes requested):
Would love to explore alternatives, if there are any, as adding 1ms would double our p75 latency <picture data-single-emoji=":dramatic-chipmunk:" title=":dramatic-chipmunk:">:dramatic-chipmunk:</picture>

dferber90 (approved):
Approving for the snapshot release as discussed. This is still missing a changeset to be release-able as a snapshot

kmk142789 (approved):

dferber90:
closing as #883 got merged in favor of this

PR #910, opened 2025-11-28
edge config snapshots

dferber90:
Adds the experimental Edge Config Snapshot feature. See [upcoming changelog](https://github.com/vercel/storage/blob/edge-config-embedding-from-postinstall/.changeset/famous-games-sleep.md) for details.

---

This PR also simplifies the test setup. We were previously invoking `jest` multiple times with different `--env` options and specific `.common.test.ts`, `.node.test.ts` and `.edge.test.ts` filters. This made it possible that some test files without those endings did not run. It also led to worse output as Jest would be invoked multiple times. I now changed the setup so we only invoke Jest once. Each env-specific test file uses the `@jest-environment` comment to specify its env. The only change to the setup is that `common.ts` tests are no longer run for both environments now run for node only.


packages/edge-config/src/create-create-client.ts
```diff
@@ -92,6 +97,48 @@ export function createCreateClient({
...
+          new Promise<TimeoutError>((resolve) => {
+            timer = setTimeout(
+              () => resolve(new TimeoutError(edgeConfigId, method, key)),
+              ms,
+            );
+          }),
+          run(),
+        ]);
+        if (result instanceof TimeoutError) throw result;
+        clearTimeout(timer);
```
AndyBitz:
```suggestion
        const result = await Promise.race([
          new Promise<TimeoutError>((resolve) => {
            timer = setTimeout(
              () => resolve(new TimeoutError(edgeConfigId, method, key)),
              ms,
            );
          }),
          run(),
        ]).finally(() => clearTimeout(timer));
        if (result instanceof TimeoutError) throw result;
```

Makes sure the timeout is always cleared when the race is done, otherwise someone could set a really high timeout value and keep the timer running, even when the `run` function throws, preventing node from exiting.

dferber90: Great catch, fixed in 03bad5b5162cbd117863760954953c293039460f

AndyBitz (approved):

packages/edge-config/src/edge-config.ts
```diff
@@ -456,4 +457,11 @@ export interface EdgeConfigClientOptions {
    * Unlike Next.js, this defaults to `no-store`, as you most likely want to use Edge Config dynamically.
    */
   cache?: 'no-store' | 'force-cache';
+
+  /**
+   * How long to wait for a fresh value before falling back to a stale value or throwing.
+   *
+   * It is recommended to only use this in combination with a bundled Edge Config (see "edge-config prepare" script).
+   */
+  timeoutMs?: number;
```
dvoytenko, line 466: The name is not clear. This is of a sort "stale while revalidate", so IMHO we should do something like `maxStaleAgeMs` and `maxWaitBeforeStaleMs`. Obv need to improve, but we need 2 options and these should be clear that they both control "stale" state. 

Also consider renaming "stale" to "snaphost".

dvoytenko (changes requested):

PR #924, opened 2026-01-09
feat(blob): Add private storage

shawnfeldman:
## Summary

Adds **private storage** support (beta), a new **`get()` method**, and **conditional gets** to `@vercel/blob`.

- **Private storage (beta)** — Upload and read private blobs with `access: 'private'` on `put()` and `get()`. Private blobs require authentication and are not publicly accessible via their URL.
- **New `get()` method** — Fetch blob content by URL or pathname. Returns a `ReadableStream` and blob metadata (url, pathname, contentType, size, etag, etc.).
- **Conditional gets** — `ifNoneMatch` option on `get()` for 304 Not Modified responses, avoiding unnecessary re-downloads.

## Example

```ts
import { put, get } from '@vercel/blob';

// Upload a private blob
const blob = await put('user123/avatar.png', file, { access: 'private' });

// Read it back
const response = await get(blob.pathname, { access: 'private' });
// response.stream — ReadableStream of the blob content
// response.blob — metadata (url, pathname, contentType, size, etag, ...)

// Conditional get — skip download if unchanged
const cached = await get(blob.pathname, {
  access: 'private',
  ifNoneMatch: response.blob.etag,
});
if (cached.statusCode === 304) {
  // Blob hasn't changed, reuse previous data
}
```

## Docs

https://vercel.com/docs/vercel-blob/private-storage


packages/blob/src/get.ts
```diff
@@ -0,0 +1,163 @@
...
+  let pathname: string;
+  let size: number | undefined;
+
+  // Check if input is a URL - if so, bypass list operation
+  if (isUrl(urlOrPathname)) {
+    blobUrl = urlOrPathname;
+    pathname = extractPathnameFromUrl(urlOrPathname);
+  } else {
+    // Use list to find the blob by pathname
+    const listResult = await list({
```
vvo: Do we have to do this? See head implementation: https://github.com/vercel/storage/blob/main/packages/blob/src/head.ts I don't think you need list.

packages/blob/src/get.ts
```diff
@@ -0,0 +1,163 @@
...
+function isUrl(urlOrPathname: string): boolean {
+  return (
+    urlOrPathname.startsWith('http://') || urlOrPathname.startsWith('https://')
+  );
+}
+
+/**
+ * Extracts the pathname from a blob URL.
+ */
+function extractPathnameFromUrl(url: string): string {
```
vvo, line 91: We do not have to do this, we must make the backend API accept both a url or pathname and it will be resolved automatically: https://github.com/vercel/storage/blob/main/packages/blob/src/head.ts#L60-L86

vvo: Sorry that's wrong, since we're not talking to the API in this case, but to the CDN. Just make sure we have the same resolution than on the backend.

packages/blob/src/get.ts
vvo: head talks to the backend API, while we won't do that here, still, you'll get content-type and content-length from the CDN response directly (http header response) so you could use that instead.

We must not do a list() call in get() because this will have billing implications.

packages/blob/src/get.ts
```diff
@@ -0,0 +1,195 @@
...
+/**
+ * Options for the get method.
+ */
+export interface GetCommandOptions extends BlobCommandOptions {
+  /**
+   * Whether the blob is publicly accessible or private.
+   * - 'public': The blob is publicly accessible via its URL.
+   * - 'private': The blob requires authentication to access.
+   */
+  access: 'public' | 'private';
```
vvo: Perhaps we could have a BlobAccess type so it can be in a single place

vvo (changes requested):
Changes requested:
- need to remove list() from get() API, use CDN response headers for size and contentType
- need to update the API of get() to `const { stream, blob } = await get(urlOrPathname)`

packages/blob/src/get.ts
shawnfeldman: adding

packages/blob/src/get.ts
```diff
@@ -0,0 +1,195 @@
...
+  let blobUrl: string;
+  let pathname: string;
+
+  // Check if input is a URL or a pathname
+  if (isUrl(urlOrPathname)) {
+    blobUrl = urlOrPathname;
+    pathname = extractPathnameFromUrl(urlOrPathname);
+  } else {
+    // Construct the URL from the token's storeId and the pathname
+    const storeId = getStoreIdFromToken(token);
```
shawnfeldman, line 178: should we use head requests instead? or extract the id off the token

packages/blob/src/get.ts
vvo: cc @shawnfeldman any comments on this? thx!

vvo (approved):

PR #954, opened 2026-01-27
[blob] feat: add useCache option to bypass CDN cache

vvo:
## Summary

Adds a `useCache` option to `blob.get()` that controls whether to fetch from CDN cache or directly from origin storage:

- `useCache: true` (default) - allows blob to be served from CDN cache
- `useCache: false` - fetches directly from origin storage

**Note:** Only effective for private blobs. Public blobs always use CDN cache.

## Why

This is useful when you need the latest version of a blob immediately after an update, bypassing any cached responses.

## Usage

```typescript
const { stream, blob } = await get('my-file.txt', {
  access: 'private',
  useCache: false, // fetch directly from origin
});
```

## Test plan

- [x] Added tests for `useCache: false` appending `?cache=0`
- [x] Added tests for `useCache: true` and omitted not appending query param
- [x] Added tests verifying returned `downloadUrl` doesn't include `cache=0`
- [x] All SDK tests passing

### DIFF
--- a/packages/blob/src/get.ts
+++ b/packages/blob/src/get.ts
@@ -15,10 +15,12 @@ export interface GetCommandOptions extends BlobCommandOptions {
   access: BlobAccessType;
   /**
    * Whether to use the content-cache layer when fetching the blob.
-   * When false, appends ?cache=0 to bypass the cache and fetch directly.
+   * When false, bypasses the cache and fetches directly from storage.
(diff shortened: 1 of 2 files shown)


packages/blob/src/get.ts
```diff
@@ -1,3 +1,4 @@
+import { fetch } from 'undici';
```
vvo:
```suggestion
```

PR #1056, opened 2026-04-20
Add Vercel OIDC auth and Presigned URLs

elliotdauber:
Adds OIDC auth for all methods in the SDK and adds support for Presigned Urls.

Adds a new `uploadPresigned` and `handleUploadPresigned` that used presigned urls for client uploads, which is needed as read-write tokens are removed.

Minor version bump.


elliotdauber:
* **#1056** <a href="https://app.graphite.com/github/pr/vercel/storage/1056?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/storage/1056?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

packages/blob/src/client.ts
```diff
@@ -307,6 +319,60 @@ export const upload = createPutMethod<UploadOptions>({
...
+
+    if (
+      // @ts-expect-error -- Runtime check for DX.
+      options.addRandomSuffix !== undefined ||
+      // @ts-expect-error -- Runtime check for DX.
+      options.createPutExtraChecks !== undefined ||
+      // @ts-expect-error -- Runtime check for DX.
+      options.cacheControlMaxAge !== undefined ||
+      // @ts-expect-error -- Runtime check for DX.
+      options.ifMatch !== undefined
```
falcoagustin, line 358: Bug: the runtime check has `createPutExtraChecks !== undefined` (which doesn't exist on `UploadOptions`) but never checks `allowOverwrite`, even though the error message below names it. Calling `uploadPresigned(file, { allowOverwrite: true })` silently slips through and gets sent to the server.

Replace `options.createPutExtraChecks` with `options.allowOverwrite`, and while you're here update `client/`upload`` in both strings to `client/`uploadPresigned`` so the thrown error names the right function.

The older `upload` at lines 296–308 has the same copy-paste — worth fixing in the same PR.

packages/blob/src/signed-token.ts
```diff
@@ -0,0 +1,558 @@
...
+ */
+function toBase64Url(base64: string): string {
+  return base64.replace(/\+/g, '-').replace(/\//g, '_').replace(/=+$/, '');
+}
+
+function normalizeStoreId(storeId: string): string {
+  const lowercase = storeId.toLowerCase();
+  return lowercase.startsWith('store_')
+    ? lowercase.slice('store_'.length)
+    : lowercase;
```
falcoagustin, line 252: This local `normalizeStoreId` lowercases the store id, but `helpers.ts:245` deliberately preserves case (with a comment: "the API is case-sensitive on this value") and the test at `api.node.test.ts:233-241` asserts `WdsHBk1w9fDO4vPW` stays as-is.

Only consumer is `publicBlobObjectUrl` above, which isn't re-exported from `index.ts` — so it's dead code today, but the moment it ships, `WdsHBk…` becomes `wdshbk…` and the URL won't resolve.

Either delete `publicBlobObjectUrl` (and this local helper) or `import { normalizeStoreId } from './helpers'` so there's a single source of truth.

packages/blob/src/helpers.ts
```diff
@@ -128,17 +159,174 @@ export interface WithUploadProgress {
...
+    // If not supplied manually, try to get storeId from the environment variable
+    const blobStoreId = readEnv('BLOB_STORE_ID');
+    if (blobStoreId) {
+      return {
+        kind: 'oidc',
+        token: oidcToken,
+        storeId: normalizeStoreId(blobStoreId),
+      };
+    }
+  }
```
falcoagustin: Silent OIDC→RW downgrade. If a caller explicitly passes `oidcToken` (a strong intent signal) but forgets `storeId`/`BLOB_STORE_ID`, this block exits without returning and control falls through to `BLOB_READ_WRITE_TOKEN` at line 301. The `oidcToken` is silently ignored.

Suggest: if `options?.oidcToken` was set and we couldn't resolve a store id, throw with something like ```oidcToken` was provided but no store id is available. Pass a `storeId` option or set `BLOB_STORE_ID`.`` The env-only path should still fall back, otherwise the auto-injected `VERCEL_OIDC_TOKEN` on Vercel would break RW-only setups.

falcoagustin (commented):
A few real bugs worth fixing before merging — inline.

packages/blob/src/signed-token.ts
elliotdauber: gonna just remove this

packages/blob/src/helpers.ts
elliotdauber: mmm yeah good idea

packages/blob/src/client.ts
elliotdauber: Fixed

falcoagustin (approved):
LGTM — all three findings from the prior round are resolved.

- `uploadPresigned.extraChecks` now correctly rejects `allowOverwrite` and the error message names the right function.
- `publicBlobObjectUrl` removed, so the case-folding footgun has no caller.
- Explicit `oidcToken` without a resolvable `storeId` now throws instead of silently downgrading to `BLOB_READ_WRITE_TOKEN`.

Minor nits to consider in a follow-up (non-blocking):
- `packages/blob/src/signed-token.ts:248-252` — the local lowercasing `normalizeStoreId` is unreferenced now; worth deleting so it can't be picked back up by accident.
- Add a test for the new `oidcToken`-without-`storeId` throw branch in `helpers.ts`.

PR #1057, opened 2026-04-23
Vercel Presigned URLs

elliotdauber:
Implements presigned urls

Requests signing token from the server, then can use it to sign urls

```ts
import {
  issueSignedToken,
  presignUrl,
} from '@vercel/blob';

const signingToken = await issueSignedToken({
  pathname: '*',
  operations: ['get']
  ttlSeconds: 10 * 60
});

const blobUrl = 'https://<storeId>.public.blob.vercel-storage.com/media/photo.png';

const presigned = await presignUrl(blobUrl, signingToken, 'GET', { ttlSeconds: 5 * 60 });

const res = await fetch(presigned);
```


packages/blob/src/client.ts
```diff
@@ -311,6 +318,60 @@ export const upload = createPutMethod<UploadOptions>({
...
+      options.addRandomSuffix !== undefined ||
+      // @ts-expect-error -- Runtime check for DX.
+      options.createPutExtraChecks !== undefined ||
+      // @ts-expect-error -- Runtime check for DX.
+      options.cacheControlMaxAge !== undefined ||
+      // @ts-expect-error -- Runtime check for DX.
+      options.ifMatch !== undefined
+    ) {
+      throw new BlobError(
+        "client/`upload` doesn't allow `addRandomSuffix`, `cacheControlMaxAge`, `allowOverwrite` or `ifMatch`. Configure these options at the server side when generating client tokens.",
```
elliotdauber, line 361: i need to update messaging here

packages/blob/src/client.ts
```diff
@@ -343,6 +404,19 @@ async function signPayload(
...
+async function verifyCallbackSignaturePresigned({
+  webhookPublicKey,
+  signature,
+  body,
+}: {
+  webhookPublicKey: string;
+  signature: string;
+  body: string;
+}): Promise<boolean> {
+  console.warn('verifyCallbackSignaturePresigned is not yet implemented');
```
elliotdauber: todo

packages/blob/src/client.ts
```diff
@@ -666,6 +775,127 @@ export async function handleUpload({
...
+  onUploadCompleted,
+}: HandleUploadPresignedOptions): Promise<
+  | { type: 'blob.generate-presigned-url'; presignedUrl: string }
+  | { type: 'blob.upload-completed'; response: 'ok' }
+> {
+  const resolvedWebhookPublicKey =
+    webhookPublicKey ?? process.env.BLOB_WEBHOOK_PUBLIC_KEY;
+
+  if (!resolvedWebhookPublicKey) {
+    throw new BlobError('Missing webhook public key');
```
elliotdauber, line 950: should we error here? or only when doing webhook callback, if onUploadCompleted is provided

packages/blob/src/helpers.ts
```diff
@@ -147,14 +162,92 @@ function readEnv(name: string): string | undefined {
...
   const [, , , storeId = ''] = token.split('_');
   return storeId;
 }
 
+/** Same query name as {@link import('./signed-token').BLOB_PRESIGN_QUERY_DELEGATION} (avoid circular import). */
+// TODO: JUST MOVE THIS
+const BLOB_PRESIGN_QUERY_DELEGATION = 'vercel-blob-delegation' as const;
+
+function normalizeDelegationStoreId(storeId: string): string {
+  const lowercase = storeId.toLowerCase();
```
elliotdauber: im not actually sure if we should be lowercasing store ids. Idk if they are case sensitive

packages/blob/src/signed-token.ts
```diff
@@ -0,0 +1,617 @@
+import { requestApi } from './api';
+import type { BlobClientTokenConstraintOptions } from './client-token-constraints';
+import { type BlobCommandOptions, BlobError, getApiUrl } from './helpers';
+
+/**
+ * Operations that may be encoded in a delegation token (e.g. read: `get` / `head`,
+ * write: `upload` for presigned control-plane writes — both single-object `PUT`
+ * ({@link controlPlaneBlobPutUrl}) and multipart `POST` ({@link controlPlaneBlobMpuUrl})).
+ */
+export type DelegationOperation = 'get' | 'head' | 'upload';
```
falcoagustin: We want 'uplaod' to be 'put'

packages/blob/src/client.ts
```diff
@@ -666,6 +844,172 @@ export async function handleUpload({
...
+ * `PUT` URL for single-object uploads, or a presigned `POST` URL for `/mpu` when multipart.
+ * Verifies upload-completed callbacks with Ed25519 (`BLOB_WEBHOOK_PUBLIC_KEY`) over `x-vercel-signature`,
+ * matching outbound `webhook_keypair` signing from api-storage when sending the callback.
+ *
+ * If `onUploadCompleted` is set, the callback target URL is resolved like {@link handleUpload}
+ * (same `VERCEL_BLOB_CALLBACK_URL` / Vercel URL rules) and passed as
+ * {@link HandleUploadPresignedIssuanceContext} to `getSignedToken` so you can include it in
+ * `issueSignedToken`.
+ */
+export async function handleUploadPresigned({
```
falcoagustin, line 933: We shouldn't export this in the client

elliotdauber: It's where handleUpload is exported so its gonna be where it's easiest for people to find when upgrading

packages/blob/src/signed-token.ts
elliotdauber: Uploads are not always puts though, mpu is post. Thats why i went with a separate 'upload' that covers both since they're the same operation. Discussed this a few days ago

packages/blob/src/client.ts
```diff
@@ -311,6 +318,60 @@ export const upload = createPutMethod<UploadOptions>({
...
+ *   - headers - (Optional) An object containing custom headers to be sent with the request to your handleUpload route. Example use-case: sending Authorization headers.
+ *   - contentType - (Optional) A string indicating the media type. By default, it's extracted from the pathname's extension.
+ *   - multipart - (Optional) Whether to use multipart upload for large files. When true, your `handleUploadPresigned` route must return a presigned `POST` URL for `/mpu` (e.g. `presignUrl(controlPlaneBlobMpuUrl(pathname), issued, 'POST')` from `@vercel/blob`; use the same `operations: ['upload']` delegation as for single-object presigned `PUT`).
+ *   - abortSignal - (Optional) AbortSignal to cancel the operation.
+ *   - onUploadProgress - (Optional) Callback to track upload progress: onUploadProgress(\{loaded: number, total: number, percentage: number\})
+ * @returns A promise that resolves to the blob information, including pathname, contentType, contentDisposition, url, and downloadUrl.
+ */
+export const uploadPresigned = createPutMethod<UploadOptions>({
```
elliotdauber, line 341: Not sure what you mean

packages/blob/src/signed-token.ts
elliotdauber: Another example: deletes are actually post requests

elliotdauber: So these "operations" are not http "methods" (though 3 of them happen to be the same names as http methods)

elliotdauber: changed to put

PR #1058, opened 2026-04-28
[release] consolidate release and release-snapshot into one workflow for Trusted Publishing

elliotdauber:
Trusted Publishing only allows one workflow. Moves release-snapshot into release, controlled by a flag


falcoagustin (approved):
LGTM. Consolidation logic is correct — `inputs.snapshot` gating works for both `push` (null → normal release) and `workflow_dispatch` paths. Updated header comment also fixes the stale version-format example from the old file.

PR #1059, opened 2026-04-30
[wip] presigned url version of handleUpload

elliotdauber:


packages/blob/src/client.ts
```diff
@@ -311,6 +318,60 @@ export const upload = createPutMethod<UploadOptions>({
...
+      options.addRandomSuffix !== undefined ||
+      // @ts-expect-error -- Runtime check for DX.
+      options.createPutExtraChecks !== undefined ||
+      // @ts-expect-error -- Runtime check for DX.
+      options.cacheControlMaxAge !== undefined ||
+      // @ts-expect-error -- Runtime check for DX.
+      options.ifMatch !== undefined
+    ) {
+      throw new BlobError(
+        "client/`upload` doesn't allow `addRandomSuffix`, `cacheControlMaxAge`, `allowOverwrite` or `ifMatch`. Configure these options at the server side when generating client tokens.",
```
elliotdauber, line 360: i need to update messaging here

packages/blob/src/client.ts
```diff
@@ -343,6 +404,19 @@ async function signPayload(
...
+async function verifyCallbackSignaturePresigned({
+  webhookPublicKey,
+  signature,
+  body,
+}: {
+  webhookPublicKey: string;
+  signature: string;
+  body: string;
+}): Promise<boolean> {
+  console.warn('verifyCallbackSignaturePresigned is not yet implemented');
```
elliotdauber, line 416: todo

packages/blob/src/client.ts
```diff
@@ -666,6 +775,130 @@ export async function handleUpload({
...
+  getSignedToken,
+  onUploadCompleted,
+}: HandleUploadPresignedOptions): Promise<
+  | { type: 'blob.generate-presigned-url'; presignedUrl: string }
+  | { type: 'blob.upload-completed'; response: 'ok' }
+> {
+  const resolvedWebhookPublicKey =
+    webhookPublicKey ?? process.env.BLOB_WEBHOOK_PUBLIC_KEY;
+  if (!resolvedWebhookPublicKey) {
+    throw new BlobError('Missing webhook public key');
```
elliotdauber, line 848: should we error here? or only when doing webhook callback, if onUploadCompleted is provided

packages/blob/src/helpers.ts
```diff
@@ -147,14 +162,92 @@ function readEnv(name: string): string | undefined {
...
   const [, , , storeId = ''] = token.split('_');
   return storeId;
 }
 
+/** Same query name as {@link import('./signed-token').BLOB_PRESIGN_QUERY_DELEGATION} (avoid circular import). */
+// TODO: JUST MOVE THIS
+const BLOB_PRESIGN_QUERY_DELEGATION = 'vercel-blob-delegation' as const;
+
+function normalizeDelegationStoreId(storeId: string): string {
+  const lowercase = storeId.toLowerCase();
```
elliotdauber, line 178: im not actually sure if we should be lowercasing store ids. Idk if they are case sensitive

elliotdauber:
merged into `elliot/presigned-urls` to make branch management easier

PR #1066, opened 2026-05-15
ci: declare workflow-level `contents: read` on 5 workflows

arpitjain099:
Pins the default `GITHUB_TOKEN` to `contents: read` on 5 workflows in `.github/workflows/` that don't call a GitHub API beyond the initial checkout.

The following files were left implicit because they reference `GITHUB_TOKEN` / use a write-scope action / trigger on `pull_request_target`. Those scopes are best declared by maintainers: `integration-tests-live.yml`.

## Why

CVE-2025-30066 (March 2025 `tj-actions/changed-files` supply-chain compromise) exfiltrated `GITHUB_TOKEN` from workflow logs. Pinning per workflow caps runtime authority irrespective of the repo or org default, gives drift protection if the default ever widens, and is credited per-file by the OpenSSF Scorecard `Token-Permissions` check.

YAML validated locally with `yaml.safe_load` on each touched file.

### DIFF
--- a/.github/workflows/deprecate-package.yml
+++ b/.github/workflows/deprecate-package.yml
@@ -23,6 +23,9 @@ on:
         required: true
         type: string
 
+permissions:
+  contents: read
+
 jobs:
   deprecate:
     name: Deprecate Package
--- a/.github/workflows/integration-tests-dev.yml
+++ b/.github/workflows/integration-tests-dev.yml
@@ -20,6 +20,9 @@ on:
   pull_request:
     branches: [main]
 
+permissions:
+  contents: read
+
 jobs:
   test:
     timeout-minutes: 60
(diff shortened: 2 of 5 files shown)


vvo (approved):
LGTM. Additive workflow-level `contents: read` hardening; matches the OpenSSF Scorecard Token-Permissions pattern. Verified each touched workflow only does checkout + build/test/deprecate and makes no GitHub API calls beyond that.

vvo:
@arpitjain099 We require signed commits: https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits can you do that? thx

arpitjain099:
> @arpitjain099 We require signed commits: https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits can you do that? thx

@vvo sure. Let me do it tonight after getting back home.

arpitjain099:
@vvo done - rebased and re-signed with SSH. Commit should now show as Verified.

Issue #816, opened 2025-01-04
Blob access control


sakarisson:
Hey Vercel team, happy new year! 

I'm starting on a new project, and I have a need to upload pictures to storage. When going over [the blob documentation](https://vercel.com/docs/storage/vercel-blob), I realized that all blobs are publicly accessible. 

The `options` param has a required `access` field, which can only be set to `public`. Is there a plan to extend this to have a private alternative?

vvo:
Hey @sakarisson, yes! We do have plans to have `private` blobs.

Here's the RFC we're sharing with interested customers, please give us feedback on it based on your usecase as "private" means different patterns for customers, thanks!

[Private_Blobs_RFC_for_customer_feedback.pdf](https://github.com/user-attachments/files/18456202/Private_Blobs_RFC_for_customer_feedback.pdf)

NotNoshy:
I like the Edge Middleware Authorization. It seems convenient for most use cases. Plus it works on Edge functions which is nice. For accessing Blobs on the server it seems a little less convenient but still reasonable (the proposed solution seems pretty easy to wrap in a function).  

I don't love Temporary URLs because it _feels_ like it could have edge cases or be misconfigured. It probably isn't a big deal - but the timeout is something everyone will have to think about before using it. I can think of cases where it would be the most useful (temporary downloads). 

For my use case I do need to access blobs from the client and rarely from the server (which probably isn't that common, but I do it).

theianzhang:
Hey @vvo thanks for sticking with this! Confirming use case #2, blob token auth, would meet my application's needs. For audit trail and compliance reasons (no ACLs), signed URLs introduce a slight business risk.

moshmage:
> Hey [@sakarisson](https://github.com/sakarisson), yes! We do have plans to have `private` blobs.
> 
> Here's the RFC we're sharing with interested customers, please give us feedback on it based on your usecase as "private" means different patterns for customers, thanks!
> 
> [Private_Blobs_RFC_for_customer_feedback.pdf](https://github.com/user-attachments/files/18456202/Private_Blobs_RFC_for_customer_feedback.pdf)

any ETA on `createPrivateBlobUrl()` ?

phyziyx:
Hi, 

I've several features on the wishlist: 
- ability to specify the max file size per upload with pre signed URL (preventing users from uploading files larger than they are permitted)
- private blobs
[...]

PR #1080, opened 2026-07-05
feat(blob): allow client uploads to set `access` (private client uploads)

### LINKED ISSUE #1079: @vercel/blob: client uploads can't set `access` — `onBeforeGenerateToken` omits it, so client uploads are locked to public
### Summary

There is no way to request **private** access for a **client upload**. `handleUpload`'s `onBeforeGenerateToken` callback — the only server-side hook where a client-upload token is minted — cannot return `access`, so every client-uploaded blob is forced to `public`, even though private Blob shipped (#816).

This is a gap left over from the private-storage rollout: server `put({ access: 'private' })` works, but the client-upload path (`upload()` → `handleUpload()` → `generateClientTokenFromReadWriteToken()`) was never given an `access` knob.

### Where it's blocked (in `packages/blob/src/client.ts`)

`onBeforeGenerateToken`'s return type is a `Pick` that omits `access`:

```ts
onBeforeGenerateToken: (
  pathname: string,
  clientPayload: string | null,
  multipart: boolean,
) => Promise<
  Pick<
    GenerateClientTokenOptions,
    | 'allowedContentTypes'
    | 'maximumSizeInBytes'
    | 'validUntil'
    | 'addRandomSuffix'
    | 'allowOverwrite'
    | 'cacheControlMaxAge'
    | 'ifMatch'
  > & { tokenPayload?: string | null; callbackUrl?: string }
>;
```

And `GenerateClientTokenOptions` has no `access` field to Pick from — it extends `BlobCommandOptions` (token/oidcToken/storeId/abortSignal) and `BlobClientTokenConstraintOptions` (size/content-types/validUntil/suffix/overwrite/cache/ifMatch). `access` lives only on `CommonCreateBlobOptions` (`packages/blob/src/helpers.ts`), which the token options do not extend.

Net effect: returning `{ access: 'private', ... }` from `onBeforeGenerateToken` is a TypeScript error, and there is no runtime channel to carry the intended access level into the signed client token.

### Reproduction

```ts
// app/api/upload/route.ts
import { handleUpload } from '@vercel/blob/client';

export async function POST(req: Request) {
  return Response.json(
    await handleUpload({
      request: req,
      body: await req.json(),
      onBeforeGenerateToken: async () => ({
        access: 'private', // ❌ TS2353: 'access' does not exist in type Pick<GenerateClientTokenOptions, ...>
        allowedContentTypes: ['application/pdf'],
      }),
    }),
  );
}
```

Confirmed on `@vercel/blob@2.5.0` (and `2.3.3`) — the `Pick` and `GenerateClientTokenOptions` shape are identical in both.

### Expected

Symmetry with the server-side API: a client upload should be able to be minted as private, e.g. `onBeforeGenerateToken` returning `{ access: 'private' }`, so the resulting blob is created with private access.

### Proposed fix (SDK side)

1. Add an optional `access?: BlobAccessType` to the client-token options (`GenerateClientTokenOptions`).
2. Add `'access'` to the `Pick` in `onBeforeGenerateToken`'s return type.

Because `generateClientTokenFromReadWriteToken` already serializes all of its non-`token` args into the signed token payload, and `handleUpload` already spreads the `onBeforeGenerateToken` result into it, `access` would flow end-to-end into the token payload with **no other runtime change**. A PR implementing exactly this (with a unit test asserting the decoded token payload carries `access`, plus a changeset) is open — see below.

### Open question for maintainers

Does the **control-plane / api-storage** side already honor an `access` field embedded in a `vercel_blob_client_*` token payload, or is additional backend work required for it to take effect? The SDK change is safe and minimal, but if the backend ignores token-level `access`, the type change alone would let developers *express* private client uploads without them actually being private — so I've flagged this explicitly rather than assume. Happy to adjust the PR (e.g. gate it, or wire it differently) based on how the backend consumes the token.

### Context

- Follow-up to #816 (private Blob storage) — that issue delivered private access for server-side operations but did not cover the client-upload token path.
- Real-world motivation: uploading receipts/documents that contain PII/PHI directly from the browser; these should be private, but client uploads currently can only be public.

kars85:
## What

Lets client uploads choose their access level. `handleUpload`'s `onBeforeGenerateToken` callback can now return `access: 'public' | 'private'`, bringing client uploads to parity with the server-side `put({ access })` API.

Fixes #1079.

## Why

Private Blob shipped in #816, but only for server-side operations. The client-upload path (`upload()` → `handleUpload()` → `generateClientTokenFromReadWriteToken()`) was never given an `access` knob, so every client-uploaded blob is forced to `public`. `onBeforeGenerateToken`'s return type is a `Pick<GenerateClientTokenOptions, …>` that omits `access`, and `GenerateClientTokenOptions` has no `access` member to pick from — so returning `{ access: 'private' }` is a TypeScript error with no runtime channel to carry the intent.

Motivating use case: uploading receipts/documents containing PII/PHI directly from the browser — these should be private, but client uploads can currently only be public.

## Change

Minimal and additive (`packages/blob/src/client.ts`):

1. Add optional `access?: BlobAccessType` to `GenerateClientTokenOptions`.
2. Add `'access'` to the `Pick` in `onBeforeGenerateToken`'s return type.

No runtime-logic change: `generateClientTokenFromReadWriteToken` already serializes all of its non-`token` args into the signed client-token payload, and `handleUpload` already spreads the `onBeforeGenerateToken` result into it — so `access` now flows end-to-end into the token payload automatically. When omitted, behavior is unchanged (public), so this is non-breaking.

## Tests

Added to `packages/blob/src/client.node.test.ts`:
- `generateClientTokenFromReadWriteToken` bakes `access` into the decoded token payload.
- The payload omits `access` when it isn't provided (unchanged default).
- `handleUpload` threads `access` from `onBeforeGenerateToken` into the minted client token.

Locally:
- `pnpm --filter @vercel/blob test:node` → 160 passed
- `pnpm --filter @vercel/blob test:edge` → passed
- `pnpm --filter @vercel/blob type-check` (`tsc --noEmit`) → clean

Changeset included (`minor`).

## Open question for maintainers 🔎

I could only verify the SDK side. **Does the control-plane / api-storage already honor an `access` field embedded in a `vercel_blob_client_*` token payload, or is backend work needed for it to take effect?** If the backend currently ignores token-level `access`, this type change would let developers *express* private client uploads without them actually being private — so I'm flagging it rather than assuming. Happy to gate it, wire it differently, or hold the SDK change until the backend consumes the token — whatever matches how you'd like this to land.

---

https://claude.ai/code/session_014yndSRPZ2WRRFJajNWX1F1


kars85:
Answering my own open question after tracing the SDK — and it changes how this PR should be read.

**Today the control-plane gets a client upload's `access` from the `x-vercel-blob-access` request header, not from the token.** In `packages/blob/src/put-helpers.ts` the access header is set unconditionally from the *client-side* option:

```ts
// access is always required, so always add it to headers
headers[putOptionHeaderMap.access] = options.access; // x-vercel-blob-access
```

So on a client upload: `upload(path, body, { access, handleUploadUrl })` → the browser sends `x-vercel-blob-access: <access>` (client-supplied) with the client token as auth, and api-storage reads access from that header. The `vercel_blob_client_*` token has never carried `access`.

**Implication:** this PR makes `access` *expressible* in `onBeforeGenerateToken` and bakes it into the signed token payload, but that field is **inert** unless api-storage is changed to read `access` from the token and enforce it over / instead of the client header. Without that backend half, setting `access` here would have no effect on the stored blob — a footgun.

I've therefore **converted this to a draft**. It's the SDK-side half of a two-part change; happy to keep it warm until (or adjust it to match) whatever the backend does. The security motivation is that a server-pinned, tamper-proof access (token) is stronger than a client-settable header — but that's a backend decision. Flagging so this doesn't merge as if the SDK change alone were sufficient.

PR #1084, opened 2026-07-07
[@vercel/blob] Add useCache option to presigned get URLs

falcoagustin:
# Problem

`get({ useCache: false })` (restored in #1081) lets bearer-token callers bypass the CDN cache, but there was no way to get the same behavior on a presigned GET URL: `presignUrl()` has no cache option, so consumers of presigned URLs had to append `cache=0` manually — relying on undocumented behavior.

# Solution

- Add `useCache?: boolean` to `PresignGetUrlOptions`. When `false`, the built URL includes `cache=0` so fetches read the latest content from origin storage (`x-vercel-cache: MISS`, `cache-control: no-store`).
- Same gating as `get()`: the CDN only supports the bypass for private blobs, so the param is skipped for `access: 'public'`.
- The param stays outside the signed canonical string, so signatures are unchanged (unit test pins this) — meaning URL holders can still add/remove the param themselves; issuers should not treat it as an enforced restriction.
- Verified end to end against a real private store: the bypass URL read fresh content from origin while the plain presigned URL kept serving the stale cached version.

### DIFF
--- a/.changeset/presigned-get-use-cache.md
+++ b/.changeset/presigned-get-use-cache.md
@@ -1,0 +1,5 @@
+---
+'@vercel/blob': minor
+---
+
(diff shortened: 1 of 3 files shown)


vvo (approved):
LGTM

PR #1088, opened 2026-07-20
Add Global Config package

AndyBitz:
- Copies `@vercel/edge-config` to `@vercel/global-config`
- Updates error messages and logs to reference the new (package) name
- Uses `GLOBAL_CONFIG` instead of `EDGE_CONFIG` (but still falls back to EDGE_CONFIG)
- Changeset config update to keep both packages in sync 

The package has been published as `@vercel/global-config@1.4.3` that would just be a drop-in replacement in name only without functional changes.

The Trusted Publishing was also set-up so this repo can publish new versions.

### DIFF
--- a/packages/global-config/src/utils/index.ts
+++ b/packages/global-config/src/utils/index.ts
@@ -1,0 +1,206 @@
+import type { Connection } from '../types';
+import { trace } from './tracing';
+
+export const ERRORS = {
+  UNAUTHORIZED: '@vercel/global-config: Unauthorized',
+  EDGE_CONFIG_NOT_FOUND: '@vercel/global-config: Global Config not found',
+};
+
+export class UnexpectedNetworkError extends Error {
+  constructor(res: Response) {
+    super(
+      `@vercel/global-config: Unexpected error due to response with status code ${res.status}`,
+    );
+  }
+}
+
+/**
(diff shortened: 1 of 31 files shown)


packages/global-config/src/utils/index.ts
```diff
@@ -0,0 +1,206 @@
...
+    const params = new URLSearchParams(text.slice(12));
+
+    const id = params.get('id');
+    const token = params.get('token');
+
+    if (!id || !token) return null;
+
+    return {
+      type: 'vercel',
+      baseUrl: `https://edge-config.vercel.com/${id}`,
```
luismeyer, line 123: should we use the global config endpoint?

AndyBitz: Good point, will update

luismeyer (commented):
should we rename all functions, variables, files?

AndyBitz:
> should we rename all functions, variables, files?

No, it's more so meant to be a drop-in replacement. If we rename everything else, we should do it thoroughly and create a new major release.

I think this should happen after `@vercel/edge-config` has been deprecated and `global-config.vercel.com` took over.

luismeyer (approved):

PR #1091, opened 2026-07-27
fix(edge-config): stale-if-error expiry check was inverted

Diwak4r:
## Summary

The stale-if-error cache expiry check in `fetch-with-cached-response.ts` was inverted, causing stale-if-error caches to \*\*never expire\*\* — they would indefinitely serve stale content regardless of how much time passed.

## Bug

The condition compared:
```ts
cachedResponseEntry.time < Date.now() + staleIfError * 1000
```

This evaluates to `true` for any positive `staleIfError` because `Date.now() + staleIfError * 1000` is always in the future. The cache entry time is always less than a future timestamp, so stale content is served forever.

## Fix

Changed the comparison (in both `handleStaleIfError` and `handleStaleIfErrorException`) to:
```ts
Date.now() < cachedResponseEntry.time + staleIfError * 1000
```

This correctly checks whether `Date.now()` has exceeded the cache entry time plus the stale-if-error window — only serving stale content when the window has not yet elapsed.

## Test

Added a regression test: after advancing time past the stale-if-error window (11s past a 10s `stale-if-error=10`), a network-fault fetch should throw — not serve stale content. The test fails before the fix and passes after.

### DIFF
--- a/.changeset/stale-if-error-expiry-fix.md
+++ b/.changeset/stale-if-error-expiry-fix.md
@@ -1,0 +1,8 @@
+---
+'@vercel/edge-config': patch
+'@vercel/global-config': patch
+---
+
+fix: correct the `stale-if-error` expiry computation so stale content is no longer served indefinitely once the configured window has passed.
+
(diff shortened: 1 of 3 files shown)


Diwak4r:
Closing this to clear out a backlog I let build up rather than leave it sitting in your queue. It was never picked up and at this age it isn't going to be. If it's still wanted, say so and I'll reopen it against current main.

PR #1092, opened 2026-07-27
feat(blob): add @vercel/blob/eve extension for Eve agents

bensabic:
Exposes the Blob SDK as an [Eve](https://eve.dev) extension at the `@vercel/blob/eve` subpath, so agents can mount Blob tools without hand-writing wrappers.

```ts
// agent/extensions/blob.ts
import blob from '@vercel/blob/eve';

export default blob({});
```

Mounts `list`, `head`, `get`, `put`, `del`, `copy`, `rename`, and `create_folder`.

## Safety and correctness

- **`del` and `rename` are gated on `approval: always()`.** Both are irreversible — `rename` copies then deletes the source — and Eve treats an omitted `approval` as `never()`. Consumers can relax this with a directory mount override, documented in the README.
- **`ctx.abortSignal` is threaded into every tool**, so cancelling a turn also cancels the in-flight Blob request.
- **`get` inlines at most 64 KiB of text.** The byte count is enforced *while reading* rather than from `content-length`, which is absent on chunked responses and would otherwise let a body of any size through into the model's context.
- **`del` and `put` reject option combinations the SDK throws on** (`ifMatch` with multiple targets; `ifMatch` with `allowOverwrite: false`) at schema validation, where the model gets an actionable message instead of a runtime error.
- **`head` reports a missing blob as `{ found: false }`** to match `get`, rather than throwing.

## Packaging

- **`zod` is an optional peer dependency (`^4`)** alongside the existing optional `eve` peer, so consumers of the core SDK don't install it.
- **`scripts/restore-package-exports.mjs` wraps `eve extension build`** and repairs the `exports` field it rewrites — it repoints `.` at the extension and injects a top-level `./tools`. The merge is driven from `package.json` itself, so adding an export needs no script change, and a failed or interrupted build restores `package.json` unchanged.
- **`CreateFolderCommandOptions` now declares `storeId` and `oidcToken`**, which were already forwarded at runtime. Type-only, backward-compatible.

## Tests

Adds 44 tests (172 → 216) covering the validation guards, `abortSignal` forwarding, `get` inline/omit branching including the `content-length` and multi-byte cases, `head` miss handling, and approval wiring.

`eve` is ESM-only and the extension uses `.js` specifiers, neither of which Jest's CJS pipeline handles. Rather than change the `jest.config.cjs` shared with `src/`, `eve/__tests__/harness.ts` bridges both locally: the `.js` specifiers resolve to the real TypeScript modules (so the SDK and HTTP path are genuinely exercised) and only the three `eve/*` entry points are shimmed.

## Verification

- `pnpm test:node` — 216 passing
- `pnpm type-check`, `pnpm check` — clean
- `pnpm build` leaves `package.json` byte-identical; verified idempotent, and verified that an interrupted build and an already-clobbered `package.json` both fail safely
- Confirmed `blob({})` type-checks against the built output and `blob()` does not

## Follow-ups (not in this PR)

- `src/get.ts` reports `size: 0` when `content-length` is absent. The tool now defends against it, but making it `null` would be a breaking change to the public `GetBlobResult` type.
- `typesVersions` hardcodes the eve subpaths, so a future new eve entry point would be picked up by `exports` but not by `typesVersions`.

### DIFF
--- a/packages/blob/eve/extension.ts
+++ b/packages/blob/eve/extension.ts
@@ -1,0 +1,25 @@
+import { defineExtension } from 'eve/extension';
+import { z } from 'zod';
+
+export default defineExtension({
+  config: z.object({
+    token: z
+      .string()
+      .optional()
+      .describe(
+        'Blob read-write token. Defaults to BLOB_READ_WRITE_TOKEN when omitted.',
+      ),
+    storeId: z
+      .string()
+      .optional()
+      .describe(
+        'Blob store id for Vercel OIDC auth. Defaults to BLOB_STORE_ID when omitted.',
+      ),
+    oidcToken: z
+      .string()
+      .optional()
+      .describe(
+        'Vercel OIDC token override. Defaults to VERCEL_OIDC_TOKEN when omitted.',
+      ),
+  }),
+});
--- a/packages/blob/eve/extension.ts
+++ b/packages/blob/eve/extension.ts
@@ -15,11 +15,5 @@ export default defineExtension({
       .describe(
         'Blob store id for Vercel OIDC auth. Defaults to BLOB_STORE_ID when omitted.',
       ),
-    oidcToken: z
-      .string()
-      .optional()
-      .describe(
-        'Vercel OIDC token override. Defaults to VERCEL_OIDC_TOKEN when omitted.',
-      ),
   }),
 });
--- a/packages/blob/eve/tools/rename.ts
+++ b/packages/blob/eve/tools/rename.ts
@@ -1,0 +1,62 @@
+import { defineTool } from 'eve/tools';
+import { always } from 'eve/tools/approval';
+import { z } from 'zod';
+
+import { rename } from '../../src/rename.js';
+import { blobCommandOptions } from '../lib/options.js';
+
+export default defineTool({
+  description:
+    'Rename or move a blob to a new pathname. Copies to the destination then deletes the source.',
+  inputSchema: z.object({
+    fromUrlOrPathname: z.string().describe('Source blob URL or pathname.'),
+    toPathname: z
+      .string()
+      .describe('Destination pathname, including file extension.'),
+    access: z
+      .enum(['public', 'private'])
+      .describe('Access for the renamed blob.'),
+    addRandomSuffix: z
+      .boolean()
+      .optional()
+      .describe(
+        'Append a random suffix to the destination pathname to avoid collisions.',
+      ),
+    allowOverwrite: z
+      .boolean()
+      .optional()
+      .describe(
+        'Allow replacing an existing blob at the destination pathname.',
+      ),
+    contentType: z
+      .string()
+      .optional()
+      .describe(
+        'Media type for the renamed blob. Defaults from the destination pathname extension.',
+      ),
+    cacheControlMaxAge: z
+      .number()
+      .int()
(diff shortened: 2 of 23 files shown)


packages/blob/eve/extension.ts
```diff
@@ -0,0 +1,25 @@
...
+      .optional()
+      .describe(
+        'Blob store id for Vercel OIDC auth. Defaults to BLOB_STORE_ID when omitted.',
+      ),
+    oidcToken: z
+      .string()
+      .optional()
+      .describe(
+        'Vercel OIDC token override. Defaults to VERCEL_OIDC_TOKEN when omitted.',
+      ),
```
ctgowrie: Don't use this in a config as it refreshes at runtime on Vercel. So it should not be provided statically

packages/blob/scripts/restore-package-exports.mjs
```diff
@@ -0,0 +1,231 @@
+/**
+ * Runs `eve extension build` and repairs the damage it does to package.json.
+ *
+ * `eve extension build` rewrites the `exports` field in place, assuming it owns
+ * the whole package: it repoints `.` at the extension entry (`./dist/eve/...`)
+ * and injects a top-level `./tools` entry. For `@vercel/blob` that is wrong --
+ * `.` and `./client` are the real package entry points, and the extension is
+ * published under the `./eve` namespace instead.
```
ctgowrie, line 8: Why does the extension not have it's own package.json?

packages/blob/eve/tools/rename.ts
```diff
@@ -0,0 +1,62 @@
+import { defineTool } from 'eve/tools';
+import { always } from 'eve/tools/approval';
+import { z } from 'zod';
+
+import { rename } from '../../src/rename.js';
+import { blobCommandOptions } from '../lib/options.js';
+
+export default defineTool({
```
ctgowrie, line 8: Not sure if this is a useful tool. I'd constrain the set a bit more

packages/blob/scripts/restore-package-exports.mjs
bensabic: Mainly to keep it discoverable from the package people already have installed, and to avoid a second package that needs its own versioning plus its own dependency on `@vercel/blob`, which can then drift across releases.

This script is the cost of that. It reads the export list from `package.json` rather than hardcoding one, restores cleanly if the build fails partway, and refuses to run against an already broken `package.json`. I tested those paths.

Happy to split it out if you'd rather, it's a contained change.

packages/blob/eve/tools/rename.ts
bensabic: Kept the full set so it maps one to one with the exported SDK methods. Anyone who knows `@vercel/blob` can guess what's there, and an unused tool only costs its description in the prompt.

`rename` also isn't quite free to compose. Doing it as `copy` then `del` means two calls, and `del` is approval gated, so a move would prompt the user twice.

packages/blob/scripts/restore-package-exports.mjs
falcoagustin: We already have this kind of behavior on the CLI if it makes sense. CLI exports some of the blob SDK functionality from another package. I agree with @ctgowrie and we can mirror that behavior

PR #1093, opened 2026-07-30
[blob] Add optimizeImage option to put and new putFromUrl method

falcoagustin:
## What

SDK support for storing optimized images:

- `put(pathname, body, { optimizeImage })` → `POST /put-optimized` (body streamed through the optimizer, optimized output stored)
- `putFromUrl(pathname, url, { access, optimizeImage })` → `POST /put-from-url` (optimizer fetches the source URL server-side)

```ts
const result = await put('avatars/foo.webp', body, {
  access: 'public',
  optimizeImage: { width: 128, quality: 75, format: 'webp' },
});

const result = await putFromUrl('avatars/foo.webp', 'https://example.com/photo.jpg', {
  access: 'public',
  optimizeImage: { width: 128, quality: 75, format: 'webp' },
});
```

## Notes

- `optimizeImage`: `{ width: number, quality?: number (default 75), format?: 'jpeg' | 'png' | 'webp' | 'avif' }` — short format names map to MIME types on the wire.
- Options are validated client-side (width integer 1–3840, quality integer 1–100, format enum), mirroring the server limits so bad values fail fast with a descriptive `BlobError` instead of a billed round trip.
- Clearly non-image sources are rejected client-side when a signal exists (`contentType` option or a `Blob`/`File` `type`); otherwise the server detects the format from the bytes.
- OIDC-only (the API rejects other token types with a clear 403).
- `optimizeImage` + `multipart` throws.
- Stored `contentType` comes from the optimizer output — and the stored pathname extension is rewritten server-side to match it (vercel/api#83451).
- Requires the corresponding Blob API support to be deployed first; do not merge until it is.
- Tests cover both new paths (`put-from-url.node.test.ts`), including validation errors, content-type rejection, `ifMatch` forwarding, and the multipart conflict.

### DIFF
--- a/packages/blob/src/put-helpers.ts
+++ b/packages/blob/src/put-helpers.ts
@@ -8,6 +8,50 @@ import type { ClientCommonCreateBlobOptions } from './client';
 import type { CommonCreateBlobOptions, PresignedUrlPayload } from './helpers';
 import { BlobError, disallowedPathnameCharacters } from './helpers';
 
+/**
+ * Options to optimize an image through Vercel Image Optimization before
+ * storing it. Requires OIDC authentication.
+ */
+export interface OptimizeImageOptions {
+  /**
+   * The desired width of the optimized image in pixels (1-3840).
+   */
+  width: number;
+  /**
+   * The desired quality of the optimized image (1-100).
+   * @defaultvalue 75
+   */
+  quality?: number;
+  /**
+   * The desired output format. The original format is preserved when omitted.
+   */
+  format?: 'jpeg' | 'png' | 'webp' | 'avif';
+}
+
+const optimizeImageFormatToMimeType = {
+  jpeg: 'image/jpeg',
+  png: 'image/png',
+  webp: 'image/webp',
+  avif: 'image/avif',
+} as const;
+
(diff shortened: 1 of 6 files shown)


packages/blob/src/put-helpers.ts
```diff
@@ -8,6 +8,100 @@ import type { ClientCommonCreateBlobOptions } from './client';
 import type { CommonCreateBlobOptions, PresignedUrlPayload } from './helpers';
 import { BlobError, disallowedPathnameCharacters } from './helpers';
 
+/**
+ * Options to optimize an image through Vercel Image Optimization before
+ * storing it. Requires OIDC authentication.
+ */
+export interface OptimizeImageOptions {
+  /**
+   * The desired width of the optimized image in pixels (1-3840).
```
styfle:
```suggestion
   * The desired width of the optimized image in pixels (1-8192).
```

https://github.com/vercel/proxy/blob/709807a66ac74de3d33773ba35e087fce8c74e9e/services-ts/image-optimization/src/server.ts#L943

styfle (approved):

PR #1095, opened 2026-08-04
fix(blob): scale multipart part size for uploads above 80 GiB

### LINKED ISSUE #1067: Multipart upload fails above 80 GB — hardcoded 8 MB part size hits 10,000 part limit
## Bug

Uploading a 135 GB file with `put()` and `multipart: true` fails at part 10,001 with:

## Root cause

`partSizeInBytes` is hardcoded to `8 * 1024 * 1024` (8 MB) in `packages/blob/src/multipart/upload.ts`. Since the Blob API enforces a maximum of 10,000 parts per upload, the effective file size ceiling is:

8 MB × 10,000 = **80 GB**

This contradicts the [changelog](https://vercel.com/changelog/5tb-file-transfers-with-vercel-blob-multipart-uploads) which states multipart uploads support files up to 5 TB.

## Environment

- `@vercel/blob` v2.3.3
- Node.js, server-side `put()` with `multipart: true`
- File: 135 GB (Protomaps planet PMTiles)

## Suggested fix

Auto-scale `partSizeInBytes` based on known file size, similar to how the AWS S3 SDK handles this:

```ts
const minPartSize = 8 * 1024 * 1024 // 8 MB
const partSizeInBytes = Math.max(minPartSize, Math.ceil(totalBytes / 9_999))

Diwak4r:
## Summary

Multipart uploads fail for files larger than 80 GiB because the part size is hardcoded to 8 MiB, and the Vercel Blob API enforces a maximum of 10,000 parts per upload (8 MiB × 10,000 = 80 GiB). This contradicts the advertised 5 TB support for multipart uploads.

The fix scales the part size up for bodies with a known size above 80 GiB so the part count always stays at or under 10,000, e.g. a 100 GiB body now uploads with ~10.2 MiB parts instead of failing at part 10,001.

## Changes

- `packages/blob/src/multipart/upload.ts`
  - Added `getPartSizeInBytes(totalToLoad)`: returns the 8 MiB default for bodies at or under 80 GiB, otherwise `Math.ceil(totalToLoad / 10_000)`.
  - `uploadAllParts` now derives the part size from `totalToLoad` instead of the module-level constant.
  - Decoupled the read-ahead memory bound (`maxBytesInMemory`) from the part size. It previously scaled with the part size (2 parts × concurrency), which would buffer gigabytes for the larger part sizes now used on big uploads; it stays at the previous 128 MiB ceiling.
- `packages/blob/src/index.node.test.ts` — new tests:
  - `getPartSizeInBytes` keeps 8 MiB at/under the 80 GiB ceiling.
  - Part size scales above 80 GiB and keeps the part count ≤ 10,000 (including for a 5 TiB body).
  - An upload above 80 GiB produces one scaled part instead of many 8 MiB parts.
  - Small bodies still split into 8 MiB parts (no regression).

Bodies with an unknown size (streams) keep the 8 MiB default, since the total length isn't known upfront.

## Tests

- `test:node` — 6 suites, 176 tests, 20 snapshots ✅
- `test:edge` — 1 suite, 1 test, 2 snapshots ✅
- `test:browser` — 3 suites, 19 tests, 3 snapshots ✅

Fixes #1067

### DIFF
--- a/packages/blob/package.json
+++ b/packages/blob/package.json
@@ -60,7 +60,7 @@
     "is-buffer": "^2.0.5",
     "is-node-process": "^1.2.0",
     "throttleit": "^2.1.0",
-    "undici": "^6.23.0"
+    "undici": "^6.27.0"
   },
   "devDependencies": {
     "@edge-runtime/jest-environment": "4.0.0",
--- a/packages/blob/src/index.node.test.ts
+++ b/packages/blob/src/index.node.test.ts
@@ -17,6 +17,7 @@ import {
   rename,
   uploadPart,
 } from './index';
+import { getPartSizeInBytes, uploadAllParts } from './multipart/upload';
 
 const BLOB_API_URL_AGENT = 'https://vercel.com';
 const BLOB_STORE_BASE_URL = 'https://storeId.public.blob.vercel-storage.com';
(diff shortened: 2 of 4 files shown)


Diwak4r:
Closing this to clear out a backlog I let build up rather than leave it sitting in your queue. It was never picked up and at this age it isn't going to be. If it's still wanted, say so and I'll reopen it against current main.

PR #1099, opened 2026-08-10
[blob] Add putImage, deprecate put optimizeImage option and putFromUrl

falcoagustin:
# Problem

Image optimization currently has two split SDK surfaces: an `optimizeImage` option on `put` (body sources) and a separate `putFromUrl` function (URL sources). Callers have to pick the right one for their source kind, and the CLI's new `blob put-image` command (vercel/vercel-internal#184) mirrors that split internally.

# Solution

Add a single `putImage(pathname, bodyOrUrl, options)` method:

- The source can be the image content itself (string, File, Blob, Buffer or Stream) or a `URL` instance pointing at a public http(s) image, which is fetched server-side (`instanceof URL` picks the path; strings are always body content, and non-http(s) URLs are rejected). Only the optimized output is stored.
- Options are shared with `put` — `access`, `addRandomSuffix`, `allowOverwrite`, `cacheControlMaxAge`, `ifMatch`, auth options — but not `contentType`: the stored content type always comes from the optimizer output and the source format is detected from the bytes, so it is excluded from the type entirely (a Blob/File .type still fails fast for non-image sources) — plus the required `optimizeImage` parameters (`width`, `quality`, `format`), mirroring `put`. `multipart` is intentionally not part of the surface (optimization cannot use it).
- `optimizeImage` is guarded at runtime so an untyped caller omitting it fails loudly instead of falling through to a regular unoptimized put.
- Deprecates the `optimizeImage` option on `put` and the `putFromUrl` function via `@deprecated` JSDoc pointing at `putImage`; both keep working unchanged.

`putImage` is a standalone implementation: it talks to `/put-optimized` and `/put-from-url` directly through the shared low-level helpers (`createPutOptions`, `createPutHeaders`, `addOptimizeImageParams`, `requestApi`) and has no dependency on the deprecated `putFromUrl` function or `put`'s `optimizeImage` branch. Tests assert the exact request paths, headers, and bodies, so wire behavior is identical to the current surfaces.


packages/blob/src/put-image.node.test.ts
```diff
@@ -0,0 +1,217 @@
...
+          headers = req.headers as Record<string, string>;
+          body = req.body as string;
+          return mockedOptimizedBlob;
+        });
+
+      await expect(
+        putImage('avatar.webp', 'image-bytes', {
+          access: 'public',
+          optimizeImage: { width: 128, quality: 80, format: 'webp' },
+        }),
```
styfle, line 54: Can we change putImage() to something like this

```js
putImage('avatar.webp', 'image-bytes', {
  access: 'public',
  width: 128,
  quality: 80,
  format: 'webp',
})
```

packages/blob/src/put-from-url.ts
```diff
@@ -45,6 +47,7 @@ export type PutFromUrlBlobResult = PutBlobResult;
  *   - storeId - (Optional) Blob store id. Used to override process.env.BLOB_STORE_ID when Vercel OIDC token is available.
  *   - abortSignal - (Optional) AbortSignal to cancel the operation.
  * @returns A promise that resolves to the stored blob information, including pathname, contentType, contentDisposition, url, and downloadUrl.
+ * @deprecated Use `putImage` instead, which accepts both a body and a URL source.
```
styfle:
```suggestion
 * @deprecated Use `putImage` instead, which accepts either a body or a URL source.
```

packages/blob/src/put-image.ts
```diff
@@ -0,0 +1,155 @@
...
+ *   - token - (Optional) A string specifying the token to use when making requests. It defaults to process.env.BLOB_READ_WRITE_TOKEN when deployed on Vercel. Ignored when Vercel OIDC token is available and either process.env.BLOB_STORE_ID or options.storeId is set.
+ *   - oidcToken - (Optional) Vercel OIDC token for authentication with `storeId` (or `BLOB_STORE_ID`); overrides `VERCEL_OIDC_TOKEN`.
+ *   - storeId - (Optional) Blob store id. Used to override process.env.BLOB_STORE_ID when Vercel OIDC token is available.
+ *   - abortSignal - (Optional) AbortSignal to cancel the operation.
+ *   - onUploadProgress - (Optional) Callback to track upload progress for body sources: onUploadProgress(\{loaded: number, total: number, percentage: number\})
+ * @returns A promise that resolves to the stored blob information, including pathname, contentType, contentDisposition, url, and downloadUrl.
+ */
+export async function putImage(
+  pathname: string,
+  bodyOrUrl: PutBody,
```
styfle:
```suggestion
  bodyOrUrl: PutBody | URL,
```

I'm thinking this can either be bytes or a proper URL type. That way we can use `instanceof URL` to differentiate which code path was intended instead of a regex.

packages/blob/src/put-image.ts
```diff
@@ -0,0 +1,155 @@
...
+   * (required, 1-8192), quality (1-100, defaults to 75) and output format
+   * (defaults to the source format).
+   */
+  optimizeImage: OptimizeImageOptions;
+}
+
+export type PutImageBlobResult = PutBlobResult;
+
+function isSourceUrl(bodyOrUrl: PutBody): bodyOrUrl is string {
+  return typeof bodyOrUrl === 'string' && /^https?:\/\//i.test(bodyOrUrl);
```
styfle: This is where we could do `instance of URL` instead of the regex.

packages/blob/src/put-image.ts
```diff
@@ -0,0 +1,155 @@
...
+  pathname: string,
+  bodyOrUrl: PutBody,
+  options: PutImageCommandOptions,
+): Promise<PutImageBlobResult> {
+  if (!options?.optimizeImage) {
+    throw new BlobError('optimizeImage is required, see usage');
+  }
+
+  if (isSourceUrl(bodyOrUrl)) {
+    if (options.contentType) {
```
styfle: Help me remember - why do we allow contentType at all?

We said in a previous PR that the content type is derived by the imgopt service so I think we should always throw when its provided, right? Or maybe warn that it will be discarded? And we can remove from the TS types for this method.

packages/blob/src/put-image.node.test.ts
falcoagustin: Done in 764ace0 — width/quality/format are now top-level options: `putImage('avatar.webp', bytes, { access: 'public', width: 128, quality: 80, format: 'webp' })`. Validation messages drop the `optimizeImage.` prefix accordingly.

packages/blob/src/put-image.ts
falcoagustin: Done in 764ace0 — the signature is now `bodyOrUrl: PutBody | URL` and the branch is `instanceof URL`; strings are always treated as body content (there's a test pinning that a URL-looking string uploads as bytes). Non-http(s) URL instances are rejected with a clear error.

packages/blob/src/put-image.ts
falcoagustin: Done in 764ace0 — regex removed, the URL path is picked with `instanceof URL`.

packages/blob/src/put-image.node.test.ts
falcoagustin: Update after discussing with @falcoagustin: we're keeping `optimizeImage` as a nested option (60f3a02) so putImage mirrors put's option shape rather than flattening width/quality/format to the top level. The `PutBody | URL` / `instanceof URL` change from the other threads stays.

packages/blob/src/put-image.ts
falcoagustin: Removed in the latest commit — `contentType` is gone from `PutImageCommandOptions` (typed as `Omit<CommonCreateBlobOptions, 'contentType'>`) and putImage no longer sends `x-content-type`. The Blob/File `.type` fail-fast stays since it doesn't need the option.

cantemizyurek (approved):

packages/blob/src/put-image.ts
```diff
@@ -0,0 +1,150 @@
...
+// content type, and the source format is detected from the bytes.
+export interface PutImageCommandOptions
+  extends Omit<CommonCreateBlobOptions, 'contentType'>,
+    WithUploadProgress {
+  /**
+   * How to optimize the image before storing it: desired width in pixels
+   * (required, 1-8192), quality (1-100, defaults to 75) and output format
+   * (defaults to the source format).
+   */
+  optimizeImage: OptimizeImageOptions;
```
styfle, line 27: This still requires a nested option

```
putImage(pathname, bodyOrUrl, {
  optimizeImage: { width, quality, format }
})
```

I thought the point of a new method was so that we could do

```
putImage(pathname, bodyOrUrl, {
  width,
  quality,
  format,
})
```

PR #1106, opened 2026-09-14
fix(blob): set public names on Blob errors

cmpadden:
## Summary

Public `BlobError` subclasses currently inherit `Error.name === "Error"`, even though callers are expected to distinguish errors such as `BlobPreconditionFailedError`. This caused a real conflict-retry loop that checked the caught error's public name to silently miss an ETag mismatch. Set each exported error's `name` to its public class name so standard error inspection, logs, and stack headers identify the failure correctly; existing `instanceof` behavior and message text remain unchanged.

The names are explicit strings rather than `new.target.name` so production minification cannot change this public diagnostic value. The assignment runs only when constructing an error and adds no work to successful requests.

## Validation

- `pnpm --filter @vercel/blob check` (passes with two pre-existing unused-code warnings)
- `pnpm --filter @vercel/blob type-check`
- `pnpm --filter @vercel/blob test:node` (8 suites, 217 tests)
- `pnpm --filter @vercel/blob build`
- Verified error names in the generated ESM and CommonJS artifacts


cmpadden:
Closing in favor of documenting the supported programmatic check: callers should use `error instanceof BlobPreconditionFailedError` rather than branch on `error.name`. Changing diagnostic names would not make name-based control flow the preferred or reliable approach.

PR #1107, opened 2026-09-18
[edge-config] Remove deprecated @vercel/edge-config package

AndyBitz:
Removes the code for `@vercel/edge-config` and `@vercel/edge-config-fs` now that the packages have been deprecated.

`@vercel/global-config` is a drop-in replacement for `@vercel/edge-config` and should be used instead.


feugy (approved):
🫡

